PR #21, opened 2020-08-18
Add .gitignore file and ignore node_modules, cache and build files

samuelsogbesan:
This PR integrates a .gitignore file, [documented](https://git-scm.com/docs/gitignore) to indicate files for git to ignore when looking for changes.

Since it purely contains local dependency files, _webapp/node_modules_ was listed in the gitignore.
Locally, dependencies can be pulled by navigating to /webapp and using the command _npm install_ 

We will also ignore webapp/.cache and webapp/dist, since they are the result of local testing with [parcel](https://parceljs.org/)
.idea is also included to remove any IDE related files.


.gitignore
```diff
@@ -0,0 +1,4 @@
+webapp/node_modules
+webapp/.cache
+webapp/dist
+.idea
```
krasimir-google, line 4: Could you please also add [target/](https://github.com/holounic/STEP/blob/master/.gitignore) which as far as I know is a folder Maven uses for build artifacts.

krasimir-google: nit: [Add a newline at the end of the file](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file). Some tools in the Unix world get confused if a nonempty text file does not end with a newline.

samuelsogbesan: Updated with commits 47515f9 and 8880282

.gitignore
```diff
@@ -0,0 +1,5 @@
...
+target/
```
krasimir-google, line 5: nit: consider sorting this list [in ASCII sort order](https://en.wikipedia.org/wiki/ASCII#Printable_characters) to ease maintenance (`.` comes before letters); along with that, add [a comment](https://stackoverflow.com/questions/8865848/comments-in-gitignore) (line starting with `#`) at the beginning along the lines of "Please keep this list in ASCII sort order".

samuelsogbesan: Updated here 4454777 and 95786b2

PR #28, opened 2020-08-19
Create structure for testing

MagoDelBlocco:
Added pom.xml dependency and file tree structure for the jUnit testing.

Includes two DELETEME files, since it didn't want to stage empty folders for push

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,13 @@
       <artifactId>gson</artifactId>
       <version>2.8.6</version>
     </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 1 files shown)


holounic:
Consider adding [Mockito](https://site.mockito.org/) dependency. It allows to use mocks, which could be very useful in our project because we are going to use database and external API. 
https://mvnrepository.com/artifact/org.mockito/mockito-core/3.5.2

holounic (changes requested):
Consider adding [mockito](https://site.mockito.org/) dependency. It allows to use mocks, which could be very useful in our project because we are going to use database and external API. 
https://mvnrepository.com/artifact/org.mockito/mockito-core/3.5.2

holounic:
It is very convenient to use other packeges for testing in Spring [a short article about testing in Spring](https://spring.io/guides/gs/testing-web/)
I would suggest to edit the current version of pom.xml and add the required dependencies to it instead of modifying the outdated version.

MagoDelBlocco:
This PR got deprecated since the migration to Spring.

PR #33, opened 2020-08-20
Spring app setup

holounic:
Create pom.xml for Spring Boot application
Complete .gitignore


pom.xml
```diff
@@ -1,53 +1,81 @@
...
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-web-services</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-webflux</artifactId>
     </dependency>
 
     <dependency>
-      <groupId>com.google.code.gson</groupId>
```
samuelsogbesan, line 31: Question - will we not rely on GSON in the future?

pom.xml
```diff
@@ -1,53 +1,81 @@
...
-      <groupId>com.google.appengine</groupId>
-      <artifactId>appengine-api-1.0-sdk</artifactId>
-      <version>1.9.59</version>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-web-services</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-webflux</artifactId>
     </dependency>
```
samuelsogbesan, line 38: Remove new line to stay consistent.

src/main/webapp/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
samuelsogbesan, line 1: Remove index.html to avoid merge conflict.

src/main/java/com/google/graphgeckos/dashboard/DashboardApplication.java
```diff
@@ -0,0 +1,13 @@
+package com.google.graphgeckos.dashboard;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
```
samuelsogbesan, line 4: [Alphabetize imports](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing). Swap ...boot.SpringApplication and ...boot.autoconfigure...

samuelsogbesan (changes requested):
A couple small changes to consider, and a question about removing the GSON dependency.

pom.xml
holounic: Spring has its own way to convert classes to GSON, so we don't need it anymore.

samuelsogbesan (approved):
LGTM

pom.xml
```diff
@@ -1,53 +1,80 @@
...
   </properties>
 
   <dependencies>
     <dependency>
-      <groupId>javax.servlet</groupId>
-      <artifactId>javax.servlet-api</artifactId>
-      <version>4.0.1</version>
-      <scope>provided</scope>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-data-rest</artifactId>
```
krasimir-google, line 24: Why do we need this? Could we maybe add this dependency later when the functionality that depends on it is added. Also for other dependencies.
c.f. `mvn spring-boot:run` runs just fine without this.

pom.xml
MagoDelBlocco: Can you please provide some source for the JSON conversion and parsing? I have not found anything related to that

src/test/java/com/google/graphgeckos/dashboard/DashboardApplicationTests.java
```diff
@@ -0,0 +1,13 @@
+package com.google.graphgeckos.dashboard;
```
krasimir-google, line 1: I believe this file doesn't really test anything. There's no behavior to unit-test as far as I can see. I'd suggest removing this.

krasimir-google (changes requested):

src/test/java/com/google/graphgeckos/dashboard/DashboardApplicationTests.java
holounic: I included it to create the project barebones structure. I think, there is no need to remove it now because we will be using it not far in the future.

.gitignore
```diff
@@ -1,6 +1,29 @@
 # Please keep this list in ASCII sort order
+!.mvn/wrapper/maven-wrapper.jar
```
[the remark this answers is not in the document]
holounic: They were added to the .gitignore automatically by [Spring Boot Initializr](https://start.spring.io/) and I decided to leave them there to avoid accidental commits

pom.xml
holounic: The list of dependencies in pom.xml is a standard list for Spring Web App generated by Spring Boot Initializr. This one, for example, is [Spring Data REST](https://spring.io/projects/spring-data-rest). This dependency is needed to access its features, not to start the app.
I think, it would be way more convenient if we could start working, having this dependencies list, and delete unused  ones after we finish project or at least decide which of them we want to use.

.gitignore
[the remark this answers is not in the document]
krasimir-google: I still don't understand their purpose.
Isn't this initializr script a thing to be run once to generate a backbone for the repo?

>I decided to leave them there to avoid accidental commits
I don't follow. Could you give an example of a situation with such accidental commits?

pom.xml
krasimir-google: There are several of these dependencies that I think make sense to stay, like:
* spring-boot-starter-web
* spring-boot-devtools
* spring-boot-maven-plugin
* spring-boot-starter-test

For the others, it's a great idea to look into them, discuss their advantages and disadvantages and decide which one to use for what, but that doesn't need them to be listed here.

I believe it would be better to introduce them along with the code that starts using them.
Deleting unused dependencies after finishing the project or after deciding which to use is really the wrong way around this: you _first_ need to have a firm understanding of which exactly you need for a component of the system _before_ you start writing the code for it.
In particular for Spring Data REST, I'm curious how are you going to incorporate its functionality into the app? What is it for?

src/test/java/com/google/graphgeckos/dashboard/DashboardApplicationTests.java
krasimir-google: Okay. I think my confusion is more about the specific `contextLoads` method:
I'm sure `contextLoads` makes sense as a test in some future PR, but right now there is no "context" that's "loaded" and the method body is empty, so this is just confusing.
How about have an empty `DashboardApplicationTests` class, or at least add a TODO to the `contextLoads` method that this is not its final form?

.gitignore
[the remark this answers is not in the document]
holounic: I looked throw them once again and it turned out that you were right. Sorry :) Will remove them

pom.xml
holounic: Makes sense to me, I'll remove them

krasimir-google (approved):
This looks great!

Thank you very much!

PR #36, opened 2020-08-21
Rest api pojo components

holounic:
Add POJO components for REST API: __BuilderInfo__ class represents the response json with aggregated buildbot data, 
__Builder__ class contains information about single buildbot, __Log__ class represents a single log of a buildbot.


src/main/java/com/google/graphgeckos/dashboard/api/components/Builder.java
```diff
@@ -3,9 +3,15 @@
...
 
+/**
+ * Represents the information about a single buildbot.
+ */
 public class Builder {
   private final String builderName;
   private final List<Log> logs;
+
+  //"failed" or "passed"
+  //TODO for mashasamsikova: add java enum representation for these states
```
samuelsogbesan, line 14: nit: add space between // and comment here and elsewhere

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -3,23 +3,29 @@
...
 public class BuildInfo {
   private final String commitHash;
   private final String timeStamp;
   private final String repository;
   private final String branch;
 
-  private final List<Builder> passedBuilders;
-  private final List<Builder> failedBuilders;
+
+  /**
```
samuelsogbesan, line 17: Delete empty comment
```suggestion
```

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -3,23 +3,29 @@
...
+   *
```
samuelsogbesan, line 18: Delete empty comment
```suggestion
```

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -3,23 +3,29 @@
...
+   */
```
samuelsogbesan, line 19: Delete empty comment
```suggestion
```

samuelsogbesan (changes requested):
Small changes (comments)

samuelsogbesan (approved):
LGTM

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,48 @@
+package com.google.graphgeckos.dashboard.api.components;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents the response json with the
```
krasimir-google, line 7: nit: these comment lines can be joined into a single one.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,48 @@
...
+  public BuildInfo(String commitHash, String timeStamp, String repository,
+                   String branch, List<Builder> builders) {
+    this.commitHash = commitHash;
+    this.timeStamp = timeStamp;
+    this.repository = repository;
+    this.branch = branch;
+    this.builders = new ArrayList<>(builders);
+  }
+
+  public String getCommitHash() {
```
krasimir-google, line 28: [please document these methods](https://google.github.io/styleguide/javaguide.html#s7.3-javadoc-where-required). Focus on the interesting bits that a reader unfamiliar with the project may be interested in: what is a commit hash  and repository? Similarly for the Builder data below.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents the response json with the
+ * aggregated buildbot data.
+ */
+public class BuildInfo {
+  private final String commitHash;
+  private final String timeStamp;
```
krasimir-google, line 12: Is there a reason why this field is a string, and not some java-native way to represent a timestamp? I'd prefer one of those here.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,48 @@
...
+  private final String repository;
+  private final String branch;
+
```
krasimir-google, line 15: nit: remove extra empty lines

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Represents the response json with the
+ * aggregated buildbot data.
+ */
+public class BuildInfo {
+  private final String commitHash;
+  private final String timeStamp;
+  private final String repository;
+  private final String branch;
```
krasimir-google, line 14: what's the distinction between repository and a branch?

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
krasimir-google, line 12: nit: [timestamp is a single word](https://en.wikipedia.org/wiki/Timestamp). Prefer `timestamp` over `timeStamp` (also in method signatures).

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
holounic: The time stamp is a sequence of numbers, I decided to use String instead of long or int because this number could be huge and even be out of range.

krasimir-google: It's surprising to me that the time stamp is a sequence of numbers. How does that sequence encode time?
Note that Java provides classes like [Date](https://docs.oracle.com/javase/7/docs/api/java/util/Date.html) and there's also other some modern libs like [joda-time](https://www.joda.org/joda-time/) that provide a Java-native timestamp class. I'd expect Spring's JSON mapping capabilities to include support for mapping JSON timestamp fields in various formats to such native Java objects, e.g., https://stackoverflow.com/questions/10286204/the-right-json-date-format

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,53 @@
+package com.google.graphgeckos.dashboard.api.components;
+
+import java.util.ArrayList;
+import java.util.List;
+
+// Represents the response json with the aggregated buildbot data.
```
krasimir-google, line 7: [Turn these comments into Javadoc-style](https://google.github.io/styleguide/javaguide.html#s7-javadoc) (starting with `/**`, ending with `*/`, for the multiline ones, each internal line starts with `*`).

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,53 @@
...
+public class BuildInfo {
+
+  // git commit hash of the tested changes commit
+  private final String commitHash;
+  /*
+   the relative time (time since the application was started),
+   allows to sort received build info in the descending order (latest goes first)
+   */
```
krasimir-google: nit: it's a bit unwieldy to mix `//`-style and `/*`-style comments in this section. I'd stick to one style, e.g., turn this section into `//`-style.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,53 @@
...
+  private final String timestamp;
```
krasimir-google: It's not clear what event this timestamp field represents: the comments says "the relative time ..." but doesn't say _of what_.
You also should describe the format of this field: format, units, etc. But note my other comment about updating the field type to more suitable Java-native data type for time.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
holounic: Thanks! I changed it to Data format, it will be very convenient especially if we decide to display it in the future.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,53 @@
+package com.google.graphgeckos.dashboard.api.components;
+
+import java.util.ArrayList;
+import java.util.List;
+
+// Represents the response json with the aggregated buildbot data.
+public class BuildInfo {
+
+  // git commit hash of the tested changes commit
```
krasimir-google, line 10: These comments are great! But you should still use Javadoc to document the associated getter methods. Clients of the class don't have direct access to the private fields and documentation on the accessors is directly available in code completion for example.
In order to not repeat too much, I'd suggest having the comments on the getters be the main source of documentation of these fields (with the most details).

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/api/components/Builder.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Represents the information about a single buildbot.
+ */
+public class Builder {
+  private final String builderName;
+  private final List<Log> logs;
+
+  // "failed" or "passed"
+  // TODO for mashasamsikova: add java enum representation for these states
+  private final String status;
```
krasimir-google, line 15: I'd suggest defining this enum right now. It does not have to be complete (can contain the "failed" and "passed" states for now). The reason is that changing this from string to an enum later, when different components start using it, would require more work.

src/main/java/com/google/graphgeckos/dashboard/api/components/Log.java
```diff
@@ -0,0 +1,23 @@
...
+public class Log {
+  private final String type;
+  private final String link;
+
+  public Log(String type, String link) {
+    this.type = type;
+    this.link = link;
+  }
+
+  public String getType() {
```
krasimir-google, line 15: please document these.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,69 @@
...
+    this.commitHash = commitHash;
+    this.timestamp = timestamp;
+    this.repository = repository;
+    this.branch = branch;
+    this.builders = new ArrayList<>(builders);
+  }
+
+
+  /**
+   * @return git commit hash of the tested changes commit
```
krasimir-google, line 35: [this is common style mistake](https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment). Update according to the tip, here and elsewhere.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Represents the response json with the aggregated buildbot data.
+ */
+public class BuildInfo {
+
+  // git commit hash
+  private final String commitHash;
+  // time of receiving build info
+  private final Date timestamp;
```
krasimir-google, line 15: I'm a bit unsure about why would we need this field with the meaning as described here (time of receiving build info). All other bits of information are useful data about the build runs, independent of this application (their source is the buildbots themselves). This however feels more like a transient property of this application (when did we query the buildbots for data), which seems odd.

This may be just me not sure about the intended purpose of this field and it may in fact be important; it's just that I'm unclear.

Note that the Buildbot API contains some timestamps like started_at and completed_at, which are inherent buildbot data: http://docs.buildbot.net/latest/developer/raml/build.html

Additionally, if the meaning of this field is is really the time of receiving the build info, consider making the name more specific, like: `received_at`, instead of `timestamp`. The reason is that the word `timestamp` is way too generic and could, by itself, be interpreted in various ways.

src/main/java/com/google/graphgeckos/dashboard/api/components/Log.java
```diff
@@ -0,0 +1,31 @@
...
+  // full log link
+  private final String link;
+
+  public Log(String type, String link) {
+    this.type = type;
+    this.link = link;
+  }
+
+  /**
+   * @return a type of the log
```
krasimir-google, line 18: A clarifying example might help make it more clear what this type is.
So far I've only seen 'stdin' log entries in examples; giving that as an example would be useful.

A suggestion to improve this (in a later PR) would be to look at other types of logs are possible and documenting them here.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuilderStatus.java
```diff
@@ -0,0 +1,25 @@
+package com.google.graphgeckos.dashboard.api.components;
+
+
+/**
+ * Represents buildbot statuses: "failed" or "passed"
```
krasimir-google, line 5: nit: end the sentence with a fullstop.

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
holounic: We decided to make this field display the time of when the tested commit was pushed

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,69 @@
...
+    this.commitHash = commitHash;
+    this.timestamp = timestamp;
+    this.repository = repository;
+    this.branch = branch;
+    this.builders = new ArrayList<>(builders);
+  }
+
+
+  /**
+   * Returns git commit hash of the tested changes commit.
```
krasimir-google, line 35: nit: capitalize git here (Git); ensure you use consistent capitalization across the file.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public BuildInfo(String commitHash, Date timestamp, String repository,
+                   String branch, List<Builder> builders) {
+    this.commitHash = commitHash;
+    this.timestamp = timestamp;
+    this.repository = repository;
+    this.branch = branch;
+    this.builders = new ArrayList<>(builders);
+  }
+
```
krasimir-google, line 32: nit: remove one empty line

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
krasimir-google: Great! As a follow-up from my previous comment, consider renaming this field to something more specific than `timestamp`, e.g., `pushed_at` could be a good option. The reason is that the word timestamp is way too generic and could, by itself, be interpreted in various ways; also this rename would allow for more consistent updates later if we add more timestamp fields to the POJO.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,69 @@
+package com.google.graphgeckos.dashboard.api.components;
```
MagoDelBlocco, line 1: Please change this and all sub-classes to com.google.graphgeckos.dashboard.storage, since it is conceptually a part of the storage component.

src/main/java/com/google/graphgeckos/dashboard/api/components/BuilderStatus.java
```diff
@@ -0,0 +1,25 @@
+package com.google.graphgeckos.dashboard.api.components;
+
```
krasimir-google, line 2: nit: remove one empty line.

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
```diff
@@ -0,0 +1,69 @@
...
+  private final Date timestamp;
+  // link to to the repository
+  private final String repository;
+  // name of the branch
+  private final String branch;
+
+  // builders
+  private final List<Builder> builders;
+
+  public BuildInfo(String commitHash, Date timestamp, String repository,
```
MagoDelBlocco, line 24: make BuildInfo constructor only package accessible. No other component besides the storage should be able to create such an object. This applies to all other subclasses' constructors

MagoDelBlocco (changes requested):

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
MagoDelBlocco, line 24: consider adding the [final](https://stackoverflow.com/questions/500508/why-should-i-use-the-keyword-final-on-a-method-parameter-in-java) keyword to all parameters which will have their instances reassigned. This applies to all other methods.

MagoDelBlocco (changes requested):

src/main/java/com/google/graphgeckos/dashboard/api/components/BuildInfo.java
holounic: Controllers and shared daos are not the parts of the storage. I would suggest to add new package with shared daos

MagoDelBlocco: The POJOs are part of the storage component, since it makes _no_ sense for them existing without the storage itself.

MagoDelBlocco (approved):
LGTM

src/main/java/com/google/graphgeckos/dashboard/storage/BuildInfo.java
```diff
@@ -0,0 +1,68 @@
...
+  // Git commit hash
+  private final String commitHash;
+  // time of the Git commit push
+  private final Date timestamp;
+  // link to to the repository
+  private final String repository;
+  // name of the branch
+  private final String branch;
+
+  // builders
```
krasimir-google, line 21: this comment is not useful; either drop it (it's documented in the getter), or expand it by a few words.

src/main/java/com/google/graphgeckos/dashboard/storage/BuildInfo.java
```diff
@@ -0,0 +1,68 @@
...
+  private final List<Builder> builders;
+
+  BuildInfo(final String commitHash, final Date timestamp, final String repository,
+                   final String branch, final List<Builder> builders) {
```
krasimir-google, line 25: nit: align the first `final` on this line with the first `final` on the previous.

krasimir-google (approved):
Good!

PR #43, opened 2020-08-24
Buildbot POJO

MagoDelBlocco:
Class containing the information coming from the Buildbot Fetchers.
It is supposed to contain the compilation info for an individual revision, from a build bot. The data included should be the builder's name, status, logs, and the commit hash of the revision which has been compiled.
The instances should be immutable, with an external (not storage) component being able to create them, and only the storage should be able to consume them.

Partially closes #12 along with #44

PR #44, opened 2020-08-24
GitHub POJO

### LINKED ISSUE #12: Create Interface between Parser and Storage Controller
Create endpoints for Parsers to POST information to datastore via Storage Controller.

Requirement Reference: 4

MagoDelBlocco:
Class containing the information coming from the GitHub Fetcher components.
It is supposed to contain the meta-data of an individual revision, specifically it's commit hash, branch and commit time.
The instances should be immutable, with an external (not storage) component being able to create them, and only the storage should be able to consume them.

Partially closes #12 along with #43


src/main/java/com/google/graphgeckos/dashboard/storage/ParsedGitData.java
```diff
@@ -0,0 +1,82 @@
...
+ * to the StorageController. Provides functionality for checking whether the information
+ * is sufficient for creating an entry. Does not modify or validity check the individual
+ * contents of each field.
+ *
+ * Fields included: 
+ * -commitHash: the commit hash of the revision built by a certain build bot.
+ * -timestamp: the formatted date when the commit was pushed.
+ * -branch: the branch of the LLVM project on which this commit was pushed.
+ */
+public class ParsedGitData {
```
samuelsogbesan, line 28: LGTM
I'd suggest a name change for the class - "GitData" rather than ParsedGitData, to better communicate the fact that this represents the "Git Data" Entity.

samuelsogbesan (approved):
LGTM

src/main/java/com/google/graphgeckos/dashboard/storage/ParsedGitData.java
```diff
@@ -0,0 +1,82 @@
...
+    this.commitHash = commitHash;
+    this.timestamp = timestamp;
+    this.branch = branch;
+  }
+
+  /**
+   * Checks for the existance of the fields necessary for creating a database entry.
+   * Does not check the validity of each field.
+   *
+   * @return True if all required fields are available, False if not.
```
Somae, line 50:
```suggestion
   * @return true if all required fields are available, false if not.
```
Reference Mike's review on #43 .

Somae (approved):
LGTM with one nit.

PR #48, opened 2020-08-25
Implement Wrapper Component

samuelsogbesan:
This PR implements a helper class used in various places to wrap related components. 
The **Wrapper** class is a simple, class-customisable div container.

It is used to abstract away the HTML div element, and structurally clean up the React Components. 

_Relies on a helper, getField, that will be integrated in a future PR. getField is a method used to get a value from an object or give a defaultValue if that value is undefined._

### DIFF
--- a/src/main/webapp/components/Wrapper.js
+++ b/src/main/webapp/components/Wrapper.js
@@ -1,0 +1,9 @@
+import * as React from "react";
+ 
+export const Wrapper = (props) => {
+  return (
+    <div>
+      {props.children}
+    </div>
+  )
+}
--- a/src/main/webapp/components/Wrapper.js
+++ b/src/main/webapp/components/Wrapper.js
@@ -1,8 +1,10 @@
 import * as React from "react";
- 
+import {getField} from "utils/getField";
+
 export const Wrapper = (props) => {
+  const className = getField(props,"className","");
   return (
-    <div>
+    <div className={className}>
       {props.children}
     </div>
   )
--- a/src/main/webapp/components/Wrapper.js
+++ b/src/main/webapp/components/Wrapper.js
@@ -7,5 +7,5 @@ export const Wrapper = (props) => {
     <div className={className}>
       {props.children}
     </div>
-  )
+  );
 }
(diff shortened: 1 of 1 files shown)


holounic (approved):
LGTM!

src/main/webapp/components/Wrapper.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+  * Container element used to wrap DOM elements.
+  * Used as a abstraction for "div" HTML element.
+  * 
+  * @param props an object specifying the child components
+  * and optionally specifying a className.
+  *
+*/
+export const Wrapper = (props) => {
```
krasimir-google, line 12: At a high level this abstraction doesn't hold its own weight: it's too generic and it's only used to set a class name. I'd suggest to directly use a <div> (or whatever's appropriate) in the concrete child components, also setting a concrete className there.
On a separate note, I'd rather define more strict conditions on props so that the code doesn't have to do default-initializations in general. Say for this component (ignoring that I think we don't need it), I'd document that props.className and props.children are required fields with their intended meanings and not use getField in that case at all.
Note that inlining this into concrete components has additional advantages, as it will allow you to specify the class name directly instead of having it part of the props dictionary. Use props to propagate data fields and try to keep style-related information out of it.

samuelsogbesan: +1 for the insight. Based on this suggestion, I will close this PR and revert to div until the purpose of this component because more precise.

samuelsogbesan:
Closing this PR after insight from @krasimir-google. 

**There are some potential drawbacks to using such a generalised wrapper**.

1. All Wrappers are now fixed as **divs**. 
Although this might be desired in some cases, it ultimately means that the functionality we build for this wrapper is only applicable in contexts where we want to wrap with a div. If we want to use a **<section>** or a **header**, etc we will no longer have access to the desired functionality. 

2. The Wrapper in its current state provides little additional utility compared to the base HTML elements.

3. We have to propagate HTML attributes and event listeners through props.

If this functionality becomes more vital in the future, this PR may be reopened.

PR #51, opened 2020-08-25
Update Main App Structure to Contain Core Components

samuelsogbesan:
This PR adds Header and BuildSnapshotContainer to the **App** component, which is the component rendered to the "root" div on the page.

This diagram below labels what the components represent:

image

### DIFF
--- a/src/main/webapp/App.js
+++ b/src/main/webapp/App.js
@@ -1,12 +1,18 @@
 import * as React from "react";
+import {Header} from "./components/Header";
+import {Wrapper} from "./components/Wrapper";
+import {BuildSnapshotContainer} from "./components/BuildSnapshotContainer";
 
 /**
- * Parent Component that is Rendered to the Page.
+ * Main App that is Rendered to the Page.
  * 
  * @param {*} props arbituary input properties for our application to use
  */
 export const App = (props) => {
   return (
-    <h1>Welcome {props.name}</h1>
+    <Wrapper>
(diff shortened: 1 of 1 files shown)


MagoDelBlocco (approved):
I do not know if this applies to either Header or BuildSnapshotContainer, but I'll leave this here: [do not close void elements](https://google.github.io/styleguide/htmlcssguide.html#Document_Type) 
`  <Wrapper>
      <Header/>
      <BuildSnapshotContainer/>
    </Wrapper>`

samuelsogbesan:
> I do not know if this applies to either Header or BuildSnapshotContainer, but I'll leave this here: [do not close void elements](https://google.github.io/styleguide/htmlcssguide.html#Document_Type)
> ` <Wrapper> <Header/> <BuildSnapshotContainer/> </Wrapper>`

@MagoDelBlocco These aren't traditional HTML elements. Instead, they are React Components that syntactically look like HTML elements with JSX's syntactic sugar. With React, it is required these components are self-closed, or else the syntax is incorrect.

MagoDelBlocco:
> 
> @MagoDelBlocco These aren't traditional HTML elements. Instead, they are React Components that syntactically look like HTML elements with JSX's syntactic sugar. With React, it is required these components are self-closed, or else the syntax is incorrect.

LGTM then

Somae (approved):
LGTM

PR #54, opened 2020-08-26
Fix master compilation failure

MagoDelBlocco:
Branches #43 and #44 generated compilation failures since they had no imports for the classes they used (namely Lists). Added import java.util.List to both ParsedBuildbotData.java and ParsedGitData.java, and changed a LinkedList to an ArrayList, for performance reasons.


holounic (approved):
LGTM!

Somae (approved):
Please consider adding to your description what exact performance implications such a change has instead of stating "for performance reasons". The more specific you are, the more knowledge you can transfer or reduce the burden on the reviewer. 

Performance is usually a trade-off, so providing additional information (we have few entries and that is why... / our data is structured like this so this kind of approach...) can help to see why doing a certain change is a win.

PR #56, opened 2020-08-26
DatastoreRevisionEntity POJO | Spring-Datastore interface

### LINKED ISSUE #52: Integrate Spring Entities with GCD
Find a way to directly add/update/delete classes with the Spring @Entity annotation.

MagoDelBlocco:
DatastoreRevisionEntity is an intermediary container, used by the DataRepository implementation to simplify create and update operations.

Uses [Spring-Google Cloud Datastore integration](https://cloud.spring.io/spring-cloud-static/Greenwich.RC1/multi/multi__spring_data_cloud_datastore.html), creating a Datastore.Entity directly from a Java class. Each property of the class has been annotated according to the Spring documentation, so that it can properly create the Entity needed inside the database.

Will be used inside the createRevisionEntry method, only for creating an entry from a ParsedGitData object and inside the updateRevisionEntry, for easily acquiring, accessing and modifying the properties of an already existent entry.

closes #52

### DIFF
--- a/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRevisionEntity.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRevisionEntity.java
@@ -41,6 +41,7 @@ class DatastoreRevisionEntity {
   @Field(name = "branch")
   private String branch;
 
+  @Unindexed
   @Field(name = "builders")
   private List<Builder> builders;
 
--- a/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRevisionEntity.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRevisionEntity.java
@@ -15,22 +15,19 @@
 package com.google.graphgeckos.dashboard.storage;
 
 import com.google.cloud.Timestamp;
-import java.util.List;
 import java.util.ArrayList;
+import java.util.List;
 
 /**
- * An immutable container used for providing stripped-down git information
- * to the GCDataRepository. Provides functionality for checking whether the information
- * is sufficient for creating an entry. Does not modify or validity check the individual
- * contents of each field.
- *
- * Fields included: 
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRevisionEntity.java
```diff
@@ -0,0 +1,109 @@
...
+import com.google.cloud.Timestamp;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An intermediary container used by the Spring - Google Cloud Datastore integration,
+ * allowing easy creation/update of new entities in the database. It is created using
+ * a commit's metadata from git, and updated using
+ * {@link com.google.graphgeckos.dashboard.storage#Builder Builder} objects, which add
+ * compilation information from each buildbot. The scope of an instance is supposed to
```
samuelsogbesan, line 26: "The scope of an instance is supposed to be contained inside a single method" - can I get a clarification on what the instance is in this context?

samuelsogbesan (approved):
LGTM

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRevisionEntity.java
MagoDelBlocco: For example, inside the createRevisionEntry method, this class is only used as follows: `DatastoreRevisionEntity newEntity = new DatastoreRevisionEntity(entryData);` followed immediately by `storage.save(newEntity);`, after which the instance's lifetime is done. Roughly the same happens for the updateRevisionEntry usecase.

MagoDelBlocco: And to directly address your question, an instance here is a constructed DatastoreRevisionEntity object.

MagoDelBlocco:
This component was made obsolete by the removal of the "index" field.

PR #59, opened 2020-08-27
Implement Structure for Header Component

### LINKED ISSUE #57: Structure Navigation Component (Header)
Add the HTML structure required for the Header Component.

samuelsogbesan:
This PR implements the components needed to structure the Header Component.

The Header content is split into two wrappers: **one containing the title content** and **one containing the navigation component**.

The content is wrapped to support the styling that will be used to justify the Title and Navigation Components to the left and right hand side of the parent container respectively.

The image below illustrates the content of the two wrappers.

image

Closes #57

### DIFF
--- a/src/main/webapp/components/Header.js
+++ b/src/main/webapp/components/Header.js
@@ -4,9 +4,14 @@ import * as React from "react";
  * Main Navigation Component
  */
 export const Header = () => {
+  const TITLE = "LLVM Build View";
+
   return (
     <header>
-    LLVM Build View
+      <Wrapper>
+        <span>{TITLE}</span>
+      </Wrapper>
+      <Navigation/>
     </header>
   );
 }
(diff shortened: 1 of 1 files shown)


MagoDelBlocco (approved):
LGTM

src/main/webapp/components/Header.js
```diff
@@ -1,12 +1,20 @@
...
  */
 export const Header = () => {
+  const TITLE = 'LLVM Build View';
+
   return (
-    <header>
-    LLVM Build View
+    <header className='header'>
+      <div>
+        <span className='title'>{TITLE}</span>
```
krasimir-google, line 13: Consider to just inline the title text directly into the span, if possible, for now. (if we need to share this text between separate bits in the future, that would be a good reason to extract it as a title.)

krasimir-google (approved):
Looks good!
Feel free to merge after resolving one suggestion I left.

src/main/webapp/components/Header.js
samuelsogbesan: Understood. I originally wrapped in a span for styling purposes. If this was inline, I'm not sure It would have the same flexibility.

PR #62, opened 2020-08-28
Implement Navigation Component Structure

samuelsogbesan:
This PR implements the **component responsible for displaying the navigation list** that is floated to the right of the header (screenshot below):

image

This component was developed with extensibility in mind, hence why the navigation options are generated from an array passed in through **props**. Each option is a JavaScript Object with the fields **link** and **text** that correspond to the **href** and **innerText** attributes of the < a > tag respectively.

### DIFF
--- a/src/main/webapp/components/Navigation.js
+++ b/src/main/webapp/components/Navigation.js
@@ -7,6 +7,8 @@ const NavigationItem = props => <li><a href={props.href}>{props.innerText}</a></
 /**
  * Main Navigation List.
  * Instance will be embedded in ./Header component.
+ * @param {options: [option]} props arbituary object containing an array of
+ * options, where each option = {href:String, text: String}
  */
 export const Navigation = (props) => {
   return (
--- a/src/main/webapp/components/Navigation.js
+++ b/src/main/webapp/components/Navigation.js
@@ -2,6 +2,7 @@ import * as React from "react";
 
 /**
   * Main List Element to be displayed on the Navigation.
+  * @param {href:String, innerText:String}
  */
 const NavigationItem = props => <li><a href={props.href}>{props.innerText}</a></li>
 
@@ -9,7 +10,7 @@ const NavigationItem = props => <li><a href={props.href}>{props.innerText}</a></
  * Main Navigation List.
  * Instance will be embedded in ./Header component.
  * @param {options: [option]} props arbituary object containing an array of
- * options, where each option = {href:String, text: String}.
+ * options, where each option = {link:String, text: String}.
  */
 export const Navigation = (props) => {
   return (
--- a/src/main/webapp/components/Navigation.js
+++ b/src/main/webapp/components/Navigation.js
@@ -1,0 +1,13 @@
+import * as React from "react";
+
+/**
+ * Main Navigation List.
+ * Instance will be embedded in ./Header component.
+ */
+export const Navigation = () => {
+  return (
+    <ul>
+
+    </ul>
+  );
+}
--- a/src/main/webapp/components/Navigation.js
+++ b/src/main/webapp/components/Navigation.js
@@ -1,6 +1,7 @@
 import * as React from "react";
 
 /**
+  * Main List Element to be displayed on the Navigation.
  */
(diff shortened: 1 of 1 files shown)


src/main/webapp/components/Navigation.js
```diff
@@ -0,0 +1,23 @@
...
+     * Main Navigation List.
+     * Instance will be embedded in ./Header component.
+     *
+     * @param {options: [option]} props arbituary object containing an array of
+     * options, where each option = {link:String, text: String}.
+     */
+    export const Navigation = (props) => {
+  return (
+    <ul>
+      {props.options.map(option => <NavigationItem href={option.link} innerText={option.text}/>)}
```
MagoDelBlocco, line 20: The indentation seems off here. I might be wrong, but I think you should ident it +4 spaces from where it is now, or the export const Navigation ... should be -4 spaces.

MagoDelBlocco (approved):

src/main/webapp/components/Navigation.js
samuelsogbesan: Seems like a side-effect from clang-format. I'll adjust it.

MagoDelBlocco: Thank you!

src/main/webapp/components/Navigation.js
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+
+/**
+ * Main List Element to be displayed on the Navigation.
+ *
+ * @param {href: a URL String to a resource, innerText: The display text}
+ */
+const NavigationItem = props =>
```
krasimir-google, line 8:
```suggestion
const NavigationItem = (props) =>
```
(to keep it consistent with `(props)` below

src/main/webapp/components/Navigation.js
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+
+/**
+ * Main List Element to be displayed on the Navigation.
+ *
+ * @param {href: a URL String to a resource, innerText: The display text}
```
krasimir-google, line 6: In jsdoc `{}` after `@param` is interpreted not as free text, but as a type tag with a specific syntax:
https://jsdoc.app/tags-param.html#names-types-and-descriptions
https://jsdoc.app/tags-type.html
```suggestion
 * @param {href: a URL String to a resource, innerText: The display text}
```
```suggestion
 * @param {href: string, innerText: string} props (documentation)
```

Since this "Option type" described here is referenced below, you could use a `@typedef` to define its properties and refer to the name of that type definition here and below. You can see some `@typedef` examples here: https://devhints.io/jsdoc

src/main/webapp/components/Navigation.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+const NavigationItem = props =>
+    <li><a href = {props.href}>{props.innerText}</a></li>
+/**
+ * Main Navigation List.
+ * Instance will be embedded in ./Header component.
+ *
+ * @param {options: [option]} props arbituary object containing an array of
```
krasimir-google, line 14: typo: arbitrary; however I'd just drop this as it's a bit redundant.
If you have a `@typedef` for an `Option` type, you can refer to it here by `{@link Option}`.

krasimir-google (changes requested):

src/main/webapp/components/Navigation.js
samuelsogbesan: Updated here c3c481f

src/main/webapp/components/Navigation.js
samuelsogbesan: Updated in relation to [Documenting properties of values in an array](https://jsdoc.app/tags-param.html#parameters-with-properties)

src/main/webapp/components/Navigation.js
samuelsogbesan: Updated in relation to [Documenting properties of values in an array](https://jsdoc.app/tags-param.html#parameters-with-properties)

krasimir-google (approved):
LGTM

Thank you!

PR #63, opened 2020-08-28
BuildInfo POJOs Spring annotations

MagoDelBlocco:
Added all needed Spring Cloud Datastore annotations for smooth GCD integration to BuildInfo and all related subclasses, and changed the timestamp field in said classes to google.cloud.Timestamp, for uniformity among all storage components.

Annotations added:
* @ Entity(name = \<kind\>) to BuildInfo and all subclasses.
* \@ Field and @ Id to BuildInfo fields.
* \@ Unindexed to builders field, since we don't need to query the database on that.
For more info:
https://cloud.spring.io/spring-cloud-static/Greenwich.RC1/multi/multi__spring_data_cloud_datastore.html

google.cloud.Timestamp:
Is the data format that can be read and transcribed into what Google Cloud Datastore uses internally. For more info: https://googleapis.dev/java/google-cloud-clients/0.92.0-alpha/com/google/cloud/datastore/TimestampValue.html

note: these files have already been reviewed in #36 . Only the Spring-specific annotations and the timestamp data type have been added on top of that.

PR #64, opened 2020-08-28
Builder Spring-Cloud integration and Documentation Rework

MagoDelBlocco:
Please refer to #63 for Spring-Cloud integration context.

Reworked the documentation of the class, so as to reflect a more scalable class, and less methods for achieving the same result.
The constructor has been rewritten, so as to directly accept a ParsedBuildbotData object, eliminating the need for the previous constructor and an additional conversion method in ParsedBuildbotData.


src/main/java/com/google/graphgeckos/dashboard/storage/ParsedBuildbotData.java
```diff
@@ -59,16 +59,6 @@ boolean validUpdateData() {
...
-   * {@link com.gogle.graphgeckos.dashboard.storage.Builder Builder}.
-   *
-   * @return a new Builder instance, with the fields copied from the calling instance.
-   */
-  Builder toBuilder() {
-    return new Builder(builderName, logs, status);
-  }
-
   /**
    * Getter for the commitHash.
```
holounic, line 63: I would suggest changing the comment to ```Returns commitHash``` and remove the ```Getter for ...``` part.

holounic (changes requested):

src/main/java/com/google/graphgeckos/dashboard/storage/ParsedBuildbotData.java
MagoDelBlocco: This isn't in the scope of this PR. I will update all the POJOs documentation to be in-line with the newer ones, once the storage component is up and running

samuelsogbesan (approved):
LGTM. Great job on the comments, very clear!

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
```diff
@@ -0,0 +1,69 @@
...
+// limitations under the License.
+
+package com.google.graphgeckos.dashboard.storage;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.springframework.cloud.gcp.data.datastore.core.mapping.Entity;
+
+/**
+ * Contains the information retrieved from a single build bot. It is used as a subclass
```
krasimir-google, line 22:
```suggestion
 * Contains the information retrieved from a single build bot. It is used as a member of BuildInfo...
```
The Builder class is not a subclass of BuildInfo.

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
```diff
@@ -0,0 +1,69 @@
...
+ * of {@link #BuildInfo BuildInfo}, and for converting ParsedBuildbotData objects
+ * to Builder objects, used by BuildInfo.
+ */
+@Entity(name = "builder")
+public class Builder {
+
+  private final String builderName;
```
krasimir-google, line 29: Consider calling this just "name" (and update related members, getters, setters, accordingly): it's implied that it is the name of this builder object.

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
```diff
@@ -0,0 +1,69 @@
...
+  // The logs of each compilation stage, stored as described by {@link #Log Log}.
+  private final List<Log> logs;
+
+  // Builder compilation status, as described by {@link #BuilderStatus BuilderStatus}.
+  private final BuilderStatus status;
+
+  /**
+   * Converts a {@link #ParsedBuildbotData ParsedBuildbotData} object to a Builder object.
+   * The fields should be sanitized before this operation, as no field in Builder may be
+   * null.
```
krasimir-google, line 40: Consider checking these conditions in this constructor; throw IllegalArgumentException if there is a non-expected field that is null and add a @throws javadoc comment about that.

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
MagoDelBlocco: This is already checked in DatastoreRepository. Should I remove explicitly checking for `validUpdateData` there, and instead check it here, and leave this exception as a fallthrough in DatastoreRepository?

krasimir-google: >Should I remove explicitly checking for validUpdateData there, and instead check it here, and leave this exception as a fallthrough in DatastoreRepository?

Yes, I think that would be good! In principle you could create an instance of this Builder class from outside of the DatastoreRepository implementation, and then it would be appropriate to explicitly check for `validUpdateData` here, in order to ensure the guarantees that this class provides (its getters should return non-nulls). 

But I'd suggest taking this a step further (in a follow-up PR): since the validity of a `ParsedBuildbotData` instance is an internal property of that class (the thing checked by `validUpdateData()`), I'd suggest to update ParsedBuildbotData's constructor to check its arguments and throw an IllegalArgumentException if the arguments would lead to producing an invalid instance. This way you ensure that all ParsedBuildbotData instances are valid, and you could eliminate the checks in client code such as here that consumes a ParsedBuildbotData instance and it expects it to be valid. Note that this will make it not possible to create invalid ParsedBuildbotData instances at runtime, but I think we don't strictly need to be able to represent such invalid instances.
Additionally, this would be better as it would identify cases where the update data is invalid earlier; this is known as fail-fast and is generally a good rule of thumb to try to fail and bring up a problem as early as it is encountered to help out with investigations.

MagoDelBlocco: Thank you for the follow-up. The same fail-fast approach would go for ParsedGitData as well, so as to avoid any further checks inside the DatastoreRepository. Both of these will be part of an upcoming PR.

MagoDelBlocco: #79 solves this issue. Should this JavaDoc be updated in any way following this change?

krasimir-google: >#79 solves this issue. Should this JavaDoc be updated in any way following this change?
Great! After #79 gets in, I'd update this sentence `The fields should be sanitized ...` to something like `"botData" must not be null.`, or just add a `@throws IllegalArgumentException if botData is null` section.

krasimir-google (approved):
Thank you! Looks good!

PR #66, opened 2020-08-28
DataRepository implementation

MagoDelBlocco:
This PR brings the DataRepository (#23) implementation to the project.

The data model used for this database is:
* primary key: commit hash
* kind: "revision" - for the entries relevant to this stage of the project
* fields:
*- timestamp
*- branch
*- builders (see #64)

This implementation relies heavily upon the API provided by the [Spring Cloud GCP](https://googleapis.dev/java/spring-cloud-gcp/1.2.2.RELEASE/index.html) framework, as such, the methods are fairly short. All inputs from the exposed methods are sanitized, to avoid undefined behaviour.

Adding/Updating entries is done with the help of the @ Entity annotated class BuildInfo (#63), as described in this [Spring Cloud GCP tutorial](https://spring.io/projects/spring-cloud-gcp#overview). All incoming classes are converted to BuildInfo, or a subclass of it, so that they can be directly added to the datastore, without any further conversions.

The getLastEntries method sorts all "revision" kind entries by timestamp, limiting the first "number" entries, starting from a given offset.


src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
```diff
@@ -0,0 +1,139 @@
...
+ * replicate such a list without querying the database for it first, and the complexity
+ * of the indexing and querying would be great, without any apparent value.
+ *
+ * Useful links:
+ * - https://googleapis.dev/java/google-cloud-datastore/latest/index.html
+ * - https://googleapis.dev/java/spring-cloud-gcp/1.2.2.RELEASE/index.html
+ * - https://cloud.google.com/datastore/docs/concepts/
+ * - https://spring.io/projects/spring-cloud-gcp#overview
+ */
+@SpringBootApplication
```
holounic, line 51: ```@SpringBootApplication``` marks the entry point of the SpringBoot app. Please consider changing it. Spring needs classes to be annotated to scan and execute them (besides POJOs, because the are just "data types" and don't execute anything).  [This article](https://www.baeldung.com/spring-component-repository-service#:~:text=Spring%20Annotations,-In%20most%20typical&text=Then%2C%20it%20registers%20each%20bean%20in%20the%20ApplicationContext.&text=%40Component%20is%20a%20generic%20stereotype,act%20as%20a%20database%20repository)  could be really helpful to choose which annotation to use in this case.

holounic (changes requested):

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
MagoDelBlocco: Thank you for the heads up! I changed it to `@Repository` in [770dbe2](https://github.com/googleinterns/step240-2020/pull/66/commits/770dbe2eec3fe72cabbc543bc47e02ad760400c1). Let me know if there is anything else to solve regarding this annotation.

holounic: Thank you!

holounic (approved):
LGTM!

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
```diff
@@ -0,0 +1,139 @@
...
+ * of the indexing and querying would be great, without any apparent value.
+ *
+ * Useful links:
+ * - https://googleapis.dev/java/google-cloud-datastore/latest/index.html
+ * - https://googleapis.dev/java/spring-cloud-gcp/1.2.2.RELEASE/index.html
+ * - https://cloud.google.com/datastore/docs/concepts/
+ * - https://spring.io/projects/spring-cloud-gcp#overview
+ */
+@Repository
+public class GCDataRepository implements DataRepository {
```
krasimir-google, line 52:
```suggestion
public class DatastoreRepository implements DataRepository {
```
The "GC" abbreviation in GCDataRepository is a bit unclear (naively I read it as "garbage-collected repository at first). I'd suggest using the name of the product (Google Datastore, or just Datastore) as a basis for the name of this class.

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.Date;
+import java.util.List;
+import org.springframework.boot.SpringApplication;
+import org.springframework.cloud.gcp.data.datastore.core.DatastoreTemplate;
+import org.springframework.stereotype.Repository;
+
+/**
+ * Implementation of the DataRepository, with the purpose of providing I/O to the Google Cloud
```
krasimir-google, line 35:
```suggestion
 * A DataRepository implementation backed up by Google Datastore.
```
(possibly followed by a link to Google Datastore docs)

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
```diff
@@ -0,0 +1,139 @@
...
+ * Datastore. Each individual database entry is modeled by the {@link #BuildInfo BuildInfo} class.
```
krasimir-google, line 36:
```suggestion
 * Datastore. Each database entry is modeled by the {@link #BuildInfo BuildInfo} class.
```
("each" together with "individual") is a bit redundant.

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
```diff
@@ -0,0 +1,139 @@
...
+ * The relevant fields for the database, which are described in that class are:
```
krasimir-google, line 37:
```suggestion
 * The relevant fields for the database are:
```
(it's implied that this documentation is about the class)

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
```diff
@@ -0,0 +1,139 @@
...
+ *    - Kind: "revision"
+ *    - Key: commit hash
+ * The database indexes after all the BuildInfo fields, besides the {@code builders} list.
```
krasimir-google, line 40:
```suggestion
 * The database indexes all BuildInfo fields, except the {@code builders} list.
```
([besides vs. except](https://english.stackexchange.com/questions/7984/difference-between-except-and-besides#:~:text=They%20have%20totally%20different%20meanings,besides%20serves%20that%20of%20inclusion.&text=The%20whole%20class%20is%20going%2C%20except%20John.&text=Besides%20the%20class%2C%20Mary%20is%20also%20going.))

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
```diff
@@ -0,0 +1,139 @@
...
+ * It is unfeasable to query after the list of builders, since it would not be possible to
```
krasimir-google, line 41: I'd omit the reason why we don't index the builders list.
Effectively, we just not need to.
This brings up an additional point: do we really need indexes over all of the remaining fields of BuildInfo? I'd suggest starting only with the indexes required to implement the current DataRepository interface; we can add another index when we need it later.

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * Useful links:
+ * - https://googleapis.dev/java/google-cloud-datastore/latest/index.html
+ * - https://googleapis.dev/java/spring-cloud-gcp/1.2.2.RELEASE/index.html
+ * - https://cloud.google.com/datastore/docs/concepts/
+ * - https://spring.io/projects/spring-cloud-gcp#overview
+ */
+@Repository
+public class GCDataRepository implements DataRepository {
+  private DatastoreTemplate storage;
```
krasimir-google, line 53: Shouldn't we have also have a constructor that initializes this?
(c.f. Item 5 from "Effective Java")

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
krasimir-google, line 52: How do we test this?

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
MagoDelBlocco: Added `@Autowired` annotation to it. Hopefully, that would auto-initialize it, given that it has a `@Component` annotation. We'll have to wait for more PRs to merge to test that at runtime. Constructing it manually proved to be very complicated, since it required the Spring context and some other data of the `@Entity` classes.

krasimir-google: Thank you! I think `@Autowired` is fine.

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
MagoDelBlocco: With #70, which will get merged in this branch

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
MagoDelBlocco: This is a valid point. Thank you for bringing this up! I will also update #63 accordingly.

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
MagoDelBlocco: removed the indexing part, according to the next comment

src/main/java/com/google/graphgeckos/dashboard/storage/GCDataRepository.java
krasimir-google: Great! I'd suggest to merge #70 into this one right away and keep working on the test cases here. Specifically, there is no need for a separate PR just for test cases. Keeping the implementation together with its tests together in a PR is one of the few exceptions to the "keep PRs small" rule.

MagoDelBlocco: In this case I will discard #70 directly, and update this PR further with the tests. Thank you for the clarification!

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  /**
+   * Tests the behaviour of add/update/get/delete requests on null data.
+   */
+  @Test
+  public void testRequestsNullData() {
+    DatastoreRepository storage = new DatastoreRepository();
+
+    Assert.assertFalse(storage.createRevisionEntry(null));
```
krasimir-google, line 84: I'd expect this to throw an `IllegalArgumentException` in this case, not return.
You can use `assertThrows` to set expectations around throwing exceptions: https://stackoverflow.com/questions/40268446/junit-5-how-to-assert-an-exception-is-thrown

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
```diff
@@ -0,0 +1,220 @@
...
+    DatastoreRepository storage = new DatastoreRepository();
+
+    Assert.assertTrue(storage.deleteRevisionEntry("2"));
+  }
+
+  /**
+   * Tests multiple threads creating update requests for the same entities.
+   */
+  @Test
+  public void testParallelBurstRequests() {
```
krasimir-google, line 123: I would not include a multithreaded unit test.
Unit tests like this are not the best way to test or set concurrency expectations. We have no control over how the Threads are going to be scheduled and how are they to interfere with each other, so effectively every run of the test could execute a very different execution. Often such tests become "flaky" (different executions pass or fail in the same configuration) which reduces their utility as a good way to prevent regressions.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
```diff
@@ -0,0 +1,156 @@
...
+   */
+  public boolean createRevisionEntry(ParsedGitData entryData) throws IllegalArgumentException {
+    if (entryData == null) {
+      throw new IllegalArgumentException("entryData cannot be null");
+    }
+
+    if (getRevisionEntry(entryData.getCommitHash()) == null) {
+      try {
+        storage.save(new BuildInfo(entryData));
+      } catch (Exception e) {
```
krasimir-google, line 61: This catch is very broad; I'd not catch it here but let it bubble up to the caller for now.
If an exception happens here, it's probably because of some connectivity or concurrence issue, so just returning `false` in that case potentially discards relevant exception information that may help with debugging. Alternatively, I'd suggest to log the exception so that we have a change to see and debug what happened.

MagoDelBlocco: the .save(T instance) method uses the DatastoreReaderWriter::put method, which by the Java Reference states:
`put
Entity put(FullEntity<?> entity)
A Datastore put (a.k.a upsert) operation: inserts an entity if it does not exist, updates it otherwise. This method will automatically allocate an id if necessary.
Parameters:
entity - the entity to put
Returns:
an Entity with the same properties and a key that is either newly allocated or the same one if key is already complete
Throws:
DatastoreException - upon failure
IllegalArgumentException - if the given entity is missing a key`

So, basically in this case we can only theoretically get a DatastoreException or an IllegalArgumentException. Since we sanitize the inputs at every stage, the IllegalArgumentException would have been thrown earlier, which leaves only the DatastoreException. If I change `Exception` to `DatastoreException`, would it make sense to leave the rest of the code as-is?

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
MagoDelBlocco: This was an oversight on my part. It should indeed throw an exception, I'll patch it with the next commit. Thank you!

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
MagoDelBlocco: How can we test concurrent use of the datastore then?

krasimir-google: >How can we test concurrent use of the datastore then?

We don't.
We rely on the way datastore handles concurrent use itself. Concurrency guarantees are properties of the code that are not well exercised by traditional testing. Typically people would review the code and check if it acts within the specifications. In a more formal setting, it's possible to employ some formal proof techniques to have a mechanically verified proof of correctness (stuff like [TLA+](https://en.wikipedia.org/wiki/TLA%2B)), but that's not practical to do at the application code level as it's a lot of work. So usually those methods are employed for the code that actually implements the database layer, and the application layer just assumes well-behaved concurrency behavior from the database. In particular, Datastore enforces [serializable isolation](https://cloud.google.com/datastore/docs/concepts/transactions#isolation_and_consistency) and the code here should be well-behaved in such context.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
krasimir-google: Great!
> If I change Exception to DatastoreException, would it make sense to leave the rest of the code as-is?

That would be OK. I'd suggest still logging the DatastoreException in the catch clause in that case so we have a chance to see the details about such problems at runtime.

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
MagoDelBlocco: > Datastore enforces serializable isolation and the code here should be well-behaved in such context.

I believe this applies to Datastore transactions, which are a separate part of the Datastore API, since there is no mention of race conditions for regular operations, like add, delete or run. I have deleted this test for now, but I think it's worth testing if everything runs as expected in a parallel environment.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
MagoDelBlocco: For now, the stack trace and exception + exception message are printed to stderr.

krasimir-google (approved):
LGTM

PR #67, opened 2020-08-31
DatastoreRepository Cleanup procedures

MagoDelBlocco:
This PR adds the cleanup procedure, parameters class and scheduling to GCDataRepository from #66 . 

The cleanup procedure is a Runnable() implementation, which queries the datastore for all entries older than a date provided by the CleanupParameters class, then deletes those entries.
The CleanupParameters class is a subclass of the GCDataRepository, which contains all the parameters needed for the cleanup procedure (e.g. Entry TTL, cleanup frequency, etc.).
The scheduling is done via the [ScheduledExecutorService](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#:~:text=Interface%20ScheduledExecutorService&text=An%20ExecutorService%20that%20can%20schedule,to%20cancel%20or%20check%20execution.) provided by Java.

### DIFF
--- a/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
@@ -14,24 +14,26 @@
 
 package com.google.graphgeckos.dashboard.storage;
 
+import com.google.cloud.datastore.DatastoreException;
 import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.StructuredQuery.OrderBy;
 import com.google.cloud.datastore.StructuredQuery.PropertyFilter;
 import com.google.cloud.datastore.TimestampValue;
-import com.google.cloud.datastore.Query;
 import com.google.cloud.Timestamp;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
-import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.SpringApplication;
 import org.springframework.cloud.gcp.data.datastore.core.DatastoreTemplate;
+import org.springframework.stereotype.Repository;
 
 /**
- * Implementation of the DataRepository, with the purpose of providing I/O to the Google Cloud
- * Datastore. Each individual database entry is modeled by the {@link #BuildInfo BuildInfo} class.
- * The relevant fields for the database, which are described in that class are:
+ * A DataRepository implementation backed up by Google Datastore.
+ * Datastore. Each database entry is modeled by the {@link #BuildInfo BuildInfo} class.
+ * The relevant fields for the database are:
  *    - Kind: "revision"
  *    - Key: commit hash
  *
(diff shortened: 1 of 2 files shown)


holounic (approved):
LGTM

Somae:
Please add tests. In this case interesting test cases would include making sure the procedure doesn't delete too many entries for example. This is good to have for the future to make sure this never regresses and relevant data is deleted.

Somae (changes requested):
Please add tests. In this case interesting test cases would include making sure the procedure doesn't delete too many entries for example. This is good to have for the future to make sure this never regresses and relevant data is deleted.

MagoDelBlocco:
> Please add tests.
This is in the scope of #70, which will get merged in #66. The order of merging should be #67 -> #70 -> #66.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
```diff
@@ -37,20 +37,59 @@
...
+      Calendar calendar = Calendar.getInstance();
+      calendar.setTime(new Date());
+      calendar.add(ttlTimeUnit, ttl);
+      return calendar.getTime();
+    }
+  };
+
+  @Autowired
   private DatastoreTemplate storage;
+  private final ScheduledExecutorService cleanScheduler = Executors.newScheduledThreadPool(1);
```
krasimir-google, line 79: Consider using Spring's `@Scheduled` [functionality](https://www.baeldung.com/spring-scheduled-tasks), more specifically its [cron expressions](https://www.baeldung.com/spring-scheduled-tasks) version for scheduling.
[Cron expressions](https://en.wikipedia.org/wiki/Cron) are a versatile, industry-standard way to specify periodic schedules and Spring's `@Scheduled` makes it easy to hook up (you just need a void method that takes no params to annotate; you don't even need to wrap it in Runnable as you need here).

An additional advantage of a cron-based approach compared to using a ScheduledExecutorService is that when we have to re-deploy or re-start the application, the cron-based approach would not us to perform cause additional cleanup calls (as ScheduledExecutorService does not remember the last time it ran if an app gets restarted or re-deployed).

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  /**
+   * Tests proper deleting/preserving a single entity stored in the datastore.
+   */
+  @Test
+  private void testSingleEntityCleanup() {
+    DatastoreRepository storage = new DatastoreRepository();
+
+    // Test deletion when entry is older than the timestamp
```
krasimir-google, line 52: I'd suggest splitting this `@Test` method up into separate `@Test` methods, each corresponding to 1 section of the code. You can base the names to the test methods on your comments.

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
```diff
@@ -0,0 +1,121 @@
...
+    BuildInfo dummy = getDummyEntity("1");
+    storage.createRevisionEntry(dummy);
+    storage.deleteEntriesOlderThan(Timestamp.now());
+    Assert.assertTrue(storage.getRevisionEntry("1") == null);
```
krasimir-google, line 56: Prefer [`assertNull`](https://junit.org/junit4/javadoc/4.8/org/junit/Assert.html#assertNull(java.lang.Object)), here and in other places; it's more specific and it prints more useful error messages
Similarly, please split up the other one

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
```diff
@@ -0,0 +1,121 @@
...
+
+    // Test preservation when the entry is created at the same time as the timestamp
+    dummy = getDummyEntity("1");
+    storage.createRevisionEntry(dummy);
+    Timestamp savedTime = dummy.getTimestamp();
+    storage.deleteEntriesOlderThan(savedTime);
+    Assert.assertFalse(storage.getRevisionEntry("1") == null);
```
krasimir-google, line 63: prefer [`assertNotNull`](https://junit.org/junit4/javadoc/4.8/org/junit/Assert.html#assertNotNull(java.lang.Object)), here and in other places

krasimir-google (commented):
Left some suggestions for the test cases.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
```diff
@@ -37,20 +34,20 @@
...
  * - https://spring.io/projects/spring-cloud-gcp#overview
  */
-@SpringBootApplication
-public class GCDataRepository implements DataRepository {
+@Repository
+public class DatastoreRepository implements DataRepository {
+  @Autowired
   private DatastoreTemplate storage;
+  private final ScheduledExecutorService cleanScheduler = Executors.newScheduledThreadPool(1);
+  private final DataRetentionParameters dataRetentionParameters;
```
krasimir-google, line 49: I believe we don't need the `cleanScheduler` and the `dataRetentionParameters` members anymore.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
```diff
@@ -133,7 +130,49 @@ public void deleteRevisionEntry(String commitHash) throws IllegalArgumentExcepti
...
    * Returns null when nothing was found.
    */
-  private BuildInfo getRevisionEntry(String commitHash) {
+  public BuildInfo getRevisionEntry(String commitHash) {
     return storage.findById(commitHash, BuildInfo.class);
   }
+
+  /**
+   * Queries all "revision" type entries, and deletes all which are older than a specified
+   * amount of time. During this operation, the GCDataRepository can be queried, but
```
krasimir-google, line 139:
```suggestion
   * amount of time. During this operation, this repository can be queried, but
```
(GCDataRepository is the obsolete name of this class)

MagoDelBlocco:
GCP has a setting for Datastore Entity lifecycles.

PR #79, opened 2020-09-01
Added fail-fast null-checking in ParsedBuildbotData and ParsedGitData

MagoDelBlocco:
Removed the self-checking methods in ParsedBuildbotData and ParsedGitData, and instead moved the checks inside their constructors. This guarantees that no instance of them can have invalid fields, and thus requires less checks further down the application.

Now, their constructors throw IllegalArgumentException when a required field is null.

### DIFF
--- a/src/main/java/com/google/graphgeckos/dashboard/storage/ParsedBuildbotData.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/storage/ParsedBuildbotData.java
@@ -58,16 +58,6 @@ public class ParsedBuildbotData {
   }
 
   /**
-   * Checks for the existance of the fields necessary for updating a database entry.
-   * Does not check the validity of each field.
-   *
-   * @return true if all required fields are available, false if not.
-   */
-  boolean validUpdateData() {
-    return (commitHash != null && builderName != null && logs != null && status != null);
-  }
-
-  /**
    * Converts this instance to an instance of
    * {@link com.gogle.graphgeckos.dashboard.storage.Builder Builder}.
    *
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/graphgeckos/dashboard/storage/ParsedBuildbotData.java
```diff
@@ -43,25 +43,20 @@
    * @param logs logs for each individual stage of compilation. See {@link
    *      #com.google.graphgeckos.dashboard.storage.Log Log}
    * @param status the results of the compilation (true for passed, false for failed)
+   * @throws IllegalArgumentException if either parameter is null
    */
   public ParsedBuildbotData(String commitHash, String builderName,
-                            List<Log> logs, BuilderStatus status) {
+                            List<Log> logs, BuilderStatus status) throws IllegalArgumentException {
+    if (commitHash != null && builderName != null && logs != null && status != null) {
```
holounic, line 50: I suppose, it should be ```(commitHash == null || builderName == null || logs == null || status == null)```

holounic (approved):
LGTM!

holounic:
I would suggest to use [```@NonNull```](https://projectlombok.org/features/NonNull) to mark the args that shouldn't be null. This can make our code easier to read.

MagoDelBlocco:
> I would suggest to use @NonNull to mark the args that shouldn't be null.

Will add that in a future PR. Thank you for the suggestion!

krasimir-google (approved):
This is great!

PR #88, opened 2020-09-03
Delete duplicated pojos

holounic:
```GitHubData``` and ```ParsedGitData```, ```Builder``` and ```ParsedBuildBotData``` were very similar. So "clones" were removed, functionalities of the remaining were extended (added annotations for json parsing).

### CODE REVIEW

File: src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java [hunk 1]
```diff
@@ -16,37 +16,93 @@
...
  */
 @Entity(name = "builder")
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class Builder {
 
+  @Transient
+  private String commitHash;
+
+  @Transient
+  private String branch;
```
MagoDelBlocco, line 38: If we do not want to store the branch, why do we need to take it at all?

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
```diff
@@ -16,37 +16,93 @@
...
+   * The logs of each compilation stage, stored as described by {@link Log}.
+   */
+  private final List<Log> logs = new ArrayList<>();
 
-  // Builder compilation status, as described by {@link #BuilderStatus BuilderStatus}.
-  private final BuilderStatus status;
+  /**
+   * Builder compilation status, as described by {@link BuilderStatus}.
+   */
+  private BuilderStatus status;
```
MagoDelBlocco, line 54: Why isn't the status field final?

MagoDelBlocco (changes requested):

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
holounic: Because in Java it is not allowed to mark fields as final if they are not assigned in the constructor.

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
holounic: Because in Java it is not allowed to mark fields as final if they are not assigned in the constructor.

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
holounic: Because it is used in the constructor of the BuildInfo class.

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
MagoDelBlocco: The question is, why don't we have the status when this is constructed.

MagoDelBlocco (approved):

src/main/java/com/google/graphgeckos/dashboard/storage/Builder.java
```diff
@@ -16,37 +16,92 @@
...
+   * Otherwise the buildbot is considered passed {@code PASSED}.
+   * @param words Words of the "text" json field.
    */
-  Builder(ParsedBuildBotData botData) throws IllegalArgumentException {
-    if (botData == null) {
-      throw new IllegalArgumentException("botData cannot be null in Builder constructor");
+  @JsonProperty("text")
+  private void extractStatus(List<String> words) {
+    for (String word : words) {
+      if (word.equals("failed")) {
```
samuelsogbesan, line 78: Suggest to use word.toLowerCase() for the comparison here.

PR #92, opened 2020-09-07
Add missing return statement to Navigation Item Component

samuelsogbesan:
This PR adds a return statement missing from the Navigation Item Component, causing the front-end to fail.


MagoDelBlocco (approved):
LGTM

krasimir-google (approved):
LGTM

PR #101, opened 2020-09-07
Deployment script and dependencies

MagoDelBlocco:
Added a script for deploying directly on the GCP. It might require gcloud authentication before running the script.
How to use: `./deploy.sh`.

Changed the deprecated dependency "maven-appengine" to "google.cloud.maven-appengine".
Added .xml file in WEB-INF for appengine to pick up and run.


deploy.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+gcloud config set project llvm-build-dashboard &&
+mvn package com.google.cloud.tools:appengine-maven-plugin:deploy
```
samuelsogbesan, line 3: (An issue needs to be created to create the app.yaml file)

samuelsogbesan (approved):
LGTM, note made to open an issue about the app.yaml file.

deploy.sh
MagoDelBlocco: #108

krasimir-google (approved):
LGTM

PR #105, opened 2020-09-08
Remove Redundant Param in App Component

samuelsogbesan:
This PR removes the props parameter from the App component since it currently expects no arguments.

An underscore (_) was used as a placeholder to indicate nothing is expected.


MagoDelBlocco (approved):
LGTM

krasimir-google (approved):
LGTM

PR #110, opened 2020-09-08
Update BuildBotData to work with DatastoreTemplate

MagoDelBlocco:
DatastoreTemplate requires it's entities to have a default constructor, properties have a field name, and setters for all fields. This PR adds all those elements for BuildBotData.

### DIFF
--- a/src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
@@ -77,43 +77,46 @@ public class BuildBotData {
   }
 
   /**
-   * Returns the name of the builder bot. Cannot be null.
+   * @return the name of the builder. Cannot be null.
    */
+  @NonNull
   public String getName() {
     return name;
   }
 
   /**
-   * Returns the list of logs for each compilation stage. Cannot be null.
+   * @return the list of logs for each compilation stage. Cannot be null.
    */
+  @NonNull
   public List<Log> getLogs() {
     return logs;
   }
 
   /**
-   * Returns the compilation status of the builder. Cannot be null.
+   * @return the compilation status of the builder. Cannot be null.
    */
+  @NonNull
   public BuilderStatus getStatus() {
     return status;
   }
 
   /**
-   * Returns the commit hash of the commit tested by the current buildbot.
+   * @return the commit hash of the commit tested by the current buildbot. Cannot be null.
    */
+  @NonNull
   public String getCommitHash() {
     return commitHash;
   }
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -95,4 +111,49 @@ public Timestamp getTimestamp() {
...
+   * Sets the commit hash of the commit tested by the current buildbot.
+   *
+   * @param commitHash the new value for the {@code commitHash} field. Should not be null.
+   */
+  public void setCommitHash(@NonNull String commitHash) {
+    this.commitHash = commitHash;
+  }
+
+  /**
+   * Sets the timestamp of the build.
```
holounic, line 151: Nit: remove this part of the comment because names of setters are self-explanatory

MagoDelBlocco: Done. Thank you!

holounic (approved):
LGTM

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -33,66 +36,119 @@
...
     this.name = name;
     this.logs.addAll(logs);
     this.status = status;
   }
 
-  public BuildBotData() {}
-
   /**
-   * Returns the name of the builder bot. Cannot be null.
+   * @return the name of the builder. Cannot be null.
```
krasimir-google, line 80: please revert this to the previous version:
https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

Javadoc comments shouldn't start with just a `@return`, or `@param`, but include a summary fragment.

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -33,66 +36,89 @@
...
-   */
+  @NonNull
   public String getName() {
     return name;
   }
 
-  /**
-   * Returns the list of logs for each compilation stage. Cannot be null.
-   */
+  @NonNull
```
krasimir-google, line 84: nit: I'd keep the comment about this getLogs from before, as it talks about logs per compilation stage, and the "compilation stage" part is not clear just from the method signature.

krasimir-google (approved):
LGTM
I've left a suggestion of bringing up one for the comments that I think might be helpful

PR #113, opened 2020-09-08
Update Log to work with DatastoreTemplate

MagoDelBlocco:
DatastoreTemplate requires it's entities to have a default constructor, properties have a field name, and setters for all fields. This PR adds all those elements for Log.

### DIFF
--- a/src/main/java/com/google/graphgeckos/dashboard/datatypes/Log.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/datatypes/Log.java
@@ -15,6 +15,7 @@
 package com.google.graphgeckos.dashboard.datatypes;
 
 import org.springframework.cloud.gcp.data.datastore.core.mapping.Entity;
+import org.springframework.cloud.gcp.data.datastore.core.mapping.Field;
 
 /**
  * Contains the log information from a single compilation stage from a given build bot.
@@ -23,10 +24,17 @@ import org.springframework.cloud.gcp.data.datastore.core.mapping.Entity;
 public class Log {
 
   // Log type (e.g "stdio")
-  private final String type;
+  @Field(name = "type")
+  private String type;
 
   // Log link (e.g "http://lab.llvm.org:8011/builders/mlir-nvidia/builds/6403/logs/stdio")
-  private final String link;
+  @Field(name = "link")
+  private String link;
+
+  /**
+   * Used only by Spring GCP.
+   */
+  public Log() { }
 
   /**
    * Constructs an instance of Log. Both type and link can be null.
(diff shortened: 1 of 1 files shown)


samuelsogbesan (approved):
LGTM

src/main/java/com/google/graphgeckos/dashboard/datatypes/Log.java
```diff
@@ -23,10 +24,17 @@
...
 
   // Log link (e.g "http://lab.llvm.org:8011/builders/mlir-nvidia/builds/6403/logs/stdio")
-  private final String link;
+  @Field(name = "link")
+  private String link;
+
+  /**
+   * Used only by Spring GCP.
+   */
+  public Log() { }
```
krasimir-google, line 37: supernit: don't use a space in an empty block: https://google.github.io/styleguide/javaguide.html#s4.1.3-braces-empty-blocks
```suggestion
  public Log() {}
```

src/main/java/com/google/graphgeckos/dashboard/datatypes/Log.java
```diff
@@ -55,4 +63,22 @@ public String getLink() {
     return link;
   }
 
+  /**
+   * Sets the log type. 
+   *
+   * @param type the new value for {@code type}. Should not be null.
+   */
+  public void setType(String type) {
```
krasimir-google, line 71: Consider adding a `@NonNull` annotation on the type and link.

MagoDelBlocco: Actually I've mistakenly copied the `Should not be null` part here. Both of these fields can be null. I updated the PR accordingly.

src/main/java/com/google/graphgeckos/dashboard/datatypes/Log.java
```diff
@@ -42,17 +50,31 @@ public Log(String[] logComponents) {
   }
 
   /**
-   * Returns the type of the log as a String. Can be null.
+   * @return the type of the log as a String. Can be null.
```
krasimir-google, line 53: nit: here and below, don't start a javadoc comment with a `@return` or `@param`; the previous version was OK: https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/datatypes/Log.java
MagoDelBlocco: but is it worth having these javadoc comments? https://google.github.io/styleguide/javaguide.html#s7.3.1-javadoc-exception-self-explanatory

krasimir-google (approved):
LGTM
Thank you!

PR #121, opened 2020-09-09
Update DataRepository interface to support buildbot indices

### LINKED ISSUE #97: Add Storage Capabilities for Build Bot Index

MagoDelBlocco:
Added two abstract methods in the DataRepository interface: getBuildbotIndex and setBuildbotIndex.
The get will simply query the database for an "index" type entry and return it's value.

The set will update the database with the new index value provided in the parameters. This method could have either been an incrementBuildbotIndex, or setBuildbotIndex. I ultimately chose the set, because it offers greater flexibility, being able to skip incrementing the index an indefinite amount of times in the case of a service loss. However, this comes at the cost of checking the `newValue` parameter, since it cannot be lower than the last known value of that index.

Closes #97


holounic (approved):
LGTM

src/main/java/com/google/graphgeckos/dashboard/storage/DataRepository.java
```diff
@@ -80,4 +81,26 @@
    * @throws IllegalArgumentException if commitHash is null
    */
   BuildInfo getRevisionEntry(String commitHash) throws IllegalArgumentException;
+
+  /**
+   * Queries the database for the last known index for a specific buildbot. Used by the fetchers.
```
Somae, line 86: "Used by the fetchers." - Documentation that describes interactions between components has the potential to go out of date (because the comment won't "break" if the interaction is removed). You could mention intent here, or remove that bit all together. Documentation that can potentially be wrong at some point can confuse future maintainers.

src/main/java/com/google/graphgeckos/dashboard/storage/DataRepository.java
```diff
@@ -80,4 +81,26 @@
...
+   *
+   * @param buildbotName the name of the buildbot to search for
+   * @return the last known revision index for which there is information in the database.
+   * @throws IllegalArgumentException if {@code buildbotName} is null
+   */
+  int getBuildbotIndex(String buildbotName) throws IllegalArgumentException, NotBoundException;
+
+  /**
+   * Updates the database with a new value for the internal buildbot revision index. Used by the
```
Somae, line 95: Same comment here as in line 86.

Somae (approved):

src/main/java/com/google/graphgeckos/dashboard/storage/DataRepository.java
```diff
@@ -17,6 +17,7 @@
 import com.google.graphgeckos.dashboard.datatypes.BuildBotData;
 import com.google.graphgeckos.dashboard.datatypes.BuildInfo;
 import com.google.graphgeckos.dashboard.datatypes.GitHubData;
+import java.rmi.NotBoundException;
```
buchgr, line 20: It's a bad practice to repurpose exceptions from packages that you are not using. The `NotBoundException` has specific semantics with regard to RMI. Instead, consider declaring your own exception types and document their semantics.

src/main/java/com/google/graphgeckos/dashboard/storage/DataRepository.java
```diff
@@ -80,4 +81,26 @@
...
   BuildInfo getRevisionEntry(String commitHash) throws IllegalArgumentException;
+
+  /**
+   * Queries the database for the last known index for a specific buildbot. Used by the fetchers.
+   *
+   * @param buildbotName the name of the buildbot to search for
+   * @return the last known revision index for which there is information in the database.
+   * @throws IllegalArgumentException if {@code buildbotName} is null
+   */
+  int getBuildbotIndex(String buildbotName) throws IllegalArgumentException, NotBoundException;
```
buchgr, line 92: Remove the `IllegalArgumentException` from the throws block. `IllegalArgumentException` is a runtime exception. Declaring it in the throws block has no meaning in Java. See https://www.geeksforgeeks.org/checked-vs-unchecked-exceptions-in-java/

Roughly speaking, you'd use checked exceptions for cases where an error is expected and your code needs to react to it. Unchecked exceptions like `IllegalArgumentException` are used to signal bugs. For example, it would be a bug in you application if `getBuildbotIndex(null)` gets called.

src/main/java/com/google/graphgeckos/dashboard/storage/DataRepository.java
```diff
@@ -80,4 +81,26 @@
...
+   * fetchers.
+   *
+   * @param buildbotName the name of the buildbot to update
+   * @param newValue the new index value
+   * @throws IllegalArgumentException if {@code buildbotName} is null
+   * @throws IndexOutOfBoundsException if newValue is lower or equal than the previous
+   *      recorded value
+   */
+  void setBuildbotIndex(String buildbotName, int newValue) throws IllegalArgumentException,
+                                                                  IndexOutOfBoundsException;
```
buchgr, line 105: Same as above. Don't use unchecked exceptions in the throws clause.

buchgr (changes requested):

src/main/java/com/google/graphgeckos/dashboard/storage/DataRepository.java
MagoDelBlocco: I'll go ahead and remove that part altogether. Thank you!

src/main/java/com/google/graphgeckos/dashboard/storage/DataRepository.java
MagoDelBlocco: In this case, I believe the code needs to react to the cases in which these exceptions would have been thrown. Even if some bug occurs where the argument is null, the server should not stop. Should this exception handling be trusted to the user, and not enforced?

src/main/java/com/google/graphgeckos/dashboard/storage/DataRepository.java
MagoDelBlocco: Addressed in #170 . Thank you!

Somae:
@MForster could you check if Jakob's comments were sufficiently addressed? Thank you!

src/main/java/com/google/graphgeckos/dashboard/storage/DataRepository.java
MForster: @MagoDelBlocco, there is no difference between having runtime exceptions in the `throws` clause or not. I think it is acceptable to have them there for documentation purposes, but considering that you also have it documented in the JavaDoc, this is of limited value.

Java servers at Google usually handle runtime exceptions by letting them bubble up the whole request and catching them there. That way the server does not die, but the request error handling needs to be done only in one place, typically by returning a 5xx error code.

MagoDelBlocco: I have removed the `throws` clause for all methods, while maintaining the JavaDoc documentation about the exceptions. Thank you for the explanation!

PR #123, opened 2020-09-09
Create BuilderIndex class to store inside Datastore

MagoDelBlocco:
Created a class to encapsulate the information needed for storing the builder index inside datastore. The primary key is the name of the builder, and it's field is the index. 

Contains all the functionality needed for Spring GCP integration (default constructor, getters and setters). Unindexed the `index` field for performance reasons.

### DIFF
--- a/src/main/java/com/google/graphgeckos/dashboard/datatypes/BuilderIndex.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/datatypes/BuilderIndex.java
@@ -16,6 +16,7 @@ package com.google.graphgeckos.dashboard.datatypes;
 
 import org.springframework.cloud.gcp.data.datastore.core.mapping.Entity;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.Field;
+import org.springframework.cloud.gcp.data.datastore.core.mapping.Unindexed;
 import org.springframework.data.annotation.Id;
 
 /**
@@ -30,6 +31,7 @@ public class BuilderIndex {
   @Field(name = "builderName")
   private String name;
 
+  @Unindexed
   @Field(name = "timestamp")
   private int index;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/graphgeckos/dashboard/datatypes/BuilderIndex.java
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * @param index the new index of the builder
+   */
+  public void setIndex(int index) {
+    this.index = index;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
samuelsogbesan, line 80: We're overriding equals here, but we aren't extending a super class / implementing an interface. Are we missing that?

MagoDelBlocco: equals is [inherited from Object](https://stackoverflow.com/questions/31019157/how-do-all-classes-inherit-from-object#:~:text=All%20classes%20inherit%20from%20java,generally)%20not%20written%20out%20anywhere.).

samuelsogbesan (approved):
LGTM

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuilderIndex.java
```diff
@@ -0,0 +1,92 @@
...
+  @Id
+  @Field(name = "builderName")
+  private String name;
+
+  @Unindexed
+  @Field(name = "timestamp")
+  private int index;
+
+  /**
+   * Used only by Spring GCP.
```
Somae, line 39: Comments that say "Used only by" are dangerous because that might change over time, but readers of the documentation might start relying on this.

Somae (approved):
LGTM except for the comment that is prone to go out of date (since documentation is not runtime / test checked).

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuilderIndex.java
MagoDelBlocco: Fixed it. Thank you!

PR #124, opened 2020-09-09
Implementation of Index set and retrieval operations on DatastoreRepository

MagoDelBlocco:
Implemented the `getBuildbotIndex` and `setBuildbotIndex` methods described by `DataRepository` in #121.

Since the user of these functions should not have to handle `BuilderIndex` (#123), the implementation for `getBuildbotIndex` had to handle the situation where there is no entry associated to the given builder name. The solution I chose was to throw a [NotBoundException](https://docs.oracle.com/javase/8/docs/api/java/rmi/NotBoundException.html), since it seemed like the most closely related predefined exception for this specific case.

Since `getBuildbotIndex` throws an exception when there is no entry associated to the given name, the elegant solution is for `setBuildbotIndex` to not use it for retrieval of the previous index value.

### DIFF
--- a/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
@@ -163,8 +163,7 @@ public class DatastoreRepository implements DataRepository {
     return storage.findById(commitHash, BuildInfo.class);
   }
 
-  public int getBuildbotIndex(String buildbotName) throws BuildbotNotFoundException,
-                                                          NullPointerException {
+  public int getBuildbotIndex(String buildbotName) {
     Preconditions.checkNotNull(buildbotName, "buildbotName cannot be null");
 
     BuilderIndex associatedEntity = storage.findById(buildbotName, BuilderIndex.class);
@@ -176,8 +175,7 @@ public class DatastoreRepository implements DataRepository {
     return associatedEntity.getIndex();
   }
 
-  public void setBuildbotIndex(String buildbotName, int newValue) throws IllegalArgumentException,
-                                                                         NullPointerException {
+  public void setBuildbotIndex(String buildbotName, int newValue) {
     Preconditions.checkNotNull(buildbotName, "buildbotName cannot be null");
 
     BuilderIndex associatedEntity = storage.findById(buildbotName, BuilderIndex.class);
@@ -192,8 +190,7 @@ public class DatastoreRepository implements DataRepository {
     storage.save(associatedEntity);
   }
 
-  public void registerNewBuildbot(String buildbotName, int value) throws IllegalArgumentException,
-                                                                         NullPointerException {
+  public void registerNewBuildbot(String buildbotName, int value) {
     Preconditions.checkNotNull(buildbotName, "buildbotName cannot be null");
 
     BuilderIndex associatedEntity = storage.findById(buildbotName, BuilderIndex.class);
--- a/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
@@ -202,7 +202,7 @@ public class DatastoreRepository implements DataRepository {
   /**
    * {@inheritDoc}
    */
-  public void registerNewBuildbot(String name, int value) {
+  public void registerNewBuildbot(String buildbotName, int value) {
     if (buildbotName == null) {
       throw new IllegalArgumentException("buildbotName cannot be null");
     }
--- a/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
@@ -163,9 +163,6 @@ public class DatastoreRepository implements DataRepository {
     return storage.findById(commitHash, BuildInfo.class);
   }
 
-  /**
-   * {@inheritDoc}
-   */
   public int getBuildbotIndex(String buildbotName) throws IllegalArgumentException, NotBoundException {
     if (buildbotName == null) {
       throw new IllegalArgumentException("buildbotName cannot be null");
@@ -180,9 +177,6 @@ public class DatastoreRepository implements DataRepository {
     return associatedEntity.getIndex();
   }
 
-  /**
-   * {@inheritDoc}
-   */
   public void setBuildbotIndex(String buildbotName, int newValue) throws IllegalArgumentException,
                                                                   IndexOutOfBoundsException {
     if (buildbotName == null) {
@@ -199,9 +193,6 @@ public class DatastoreRepository implements DataRepository {
     storage.save(associatedEntity);
   }
 
-  /**
-   * {@inheritDoc}
-   */
   public void registerNewBuildbot(String buildbotName, int value) {
     if (buildbotName == null) {
       throw new IllegalArgumentException("buildbotName cannot be null");
--- a/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
@@ -166,7 +166,7 @@ public class DatastoreRepository implements DataRepository {
   /**
    * {@inheritDoc}
    */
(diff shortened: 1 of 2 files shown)


holounic (approved):
LGTM

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
```diff
@@ -209,4 +218,33 @@ public void testGetLastRevisionsQueryAfterDeletion() throws IOException, Interru
...
+  }
+
+  @Test
+  public void testInvalidSetIndexRequest() throws IOException, InterruptedException {
+    DatastoreRepository storage = new DatastoreRepository(emulator.getOptions().getService());
+
+    storage.setBuildbotIndex("tester", 1000);
+
+    Assert.assertThrows(IndexOutOfBoundsException.class, () -> {
+      storage.setBuildbotIndex("tester", 1);
```
Somae, line 247: Maybe add a very short comment here why it's invalid.

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
```diff
@@ -209,4 +218,33 @@ public void testGetLastRevisionsQueryAfterDeletion() throws IOException, Interru
     Assert.assertEquals(results.get(0), dummy2);
     Assert.assertEquals(results.get(1), dummy1);
   }
+
+  @Test
+  public void testValidGetAndSetIndexRequests() throws IOException, InterruptedException {
```
Somae, line 223: This tests uses setBuildbotIndex and getBuildbotIndex at once. It would be better to test them separately and inspect / prepare the emulator instead of using both methods. This allows for isolated testing. Please consider this in case the emulator allows for this.

Testing at least with 2 different value pairs shows that you can use different values / store multiple things. You could for example call the the setter and getter multiple times in the same test.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
```diff
@@ -160,4 +162,40 @@ public BuildInfo getRevisionEntry(String commitHash) throws IllegalArgumentExcep
...
+  void setBuildbotIndex(String buildbotName, int newValue) throws IllegalArgumentException,
+                                                                  IndexOutOfBoundsException {
+    if (buildbotName == null) {
+      throw new IllegalArgumentException("buildbotName cannot be null");
+    }
+
+    BuilderIndex associatedEntity = storage.findById(buildbotName, BuilderIndex.class);
+
+    if (associatedEntity == null || newValue <= associatedEntity.getIndex()) {
+      throw new IndexOutOfBoundsException("newValue cannot be lower than the previous index");
```
Somae, line 195:
```suggestion
      throw new IndexOutOfBoundsException("newValue cannot be lower than or equal the previous index");
```

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
```diff
@@ -209,4 +218,33 @@ public void testGetLastRevisionsQueryAfterDeletion() throws IOException, Interru
...
+    DatastoreRepository storage = new DatastoreRepository(emulator.getOptions().getService());
+
+    Assert.assertThrows(NotBoundException.class, () -> {
+      storage.getBuildbotIndex("tester");
+    });
+  }
+
+  @Test
+  public void testInvalidSetIndexRequest() throws IOException, InterruptedException {
+    DatastoreRepository storage = new DatastoreRepository(emulator.getOptions().getService());
```
Somae, line 242: Consider adding a line here that shows that equal index leads to an exception as well.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
```diff
@@ -160,4 +162,40 @@ public BuildInfo getRevisionEntry(String commitHash) throws IllegalArgumentExcep
...
+      throw new NotBoundException("buildbotName not bound to any index");
+    }
+
+    return associatedEntity.getIndex();
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  void setBuildbotIndex(String buildbotName, int newValue) throws IllegalArgumentException,
```
Somae, line 186: You are testing only with positive values. Are negative indices accepted as well? If yes, please add a test showing that it's valid. The documentation in pull request #121 doesn't speak to this either.

Somae (changes requested):

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
MagoDelBlocco: The emulator allows only for blank states at the beginning of each test. That being said, I could directly use the instance of Datastore that I provide to the DatastoreRepository to create the entities I need. Should I apply this to all other tests, since many of them depend on multiple methods at the same time?

Somae: Yes, I think that would be good! The more we can isolate the newly tested functionality, the better.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
MagoDelBlocco: While negative indices should be caught under the "not <= than the previous value", I added this condition to throw Index OOB in #121's documentation, since this could happen without triggering the exception for the first creation of an entry.

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
MagoDelBlocco: Done, thank you!

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
MagoDelBlocco: My plan is to add documentation to the whole Tests class in a PR, since many tests are missing explanations

Somae (approved):

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
```diff
@@ -160,4 +162,63 @@ public BuildInfo getRevisionEntry(String commitHash) throws IllegalArgumentExcep
...
+    if (associatedEntity == null) {
+      if (value < 0) {
+        throw new IndexOutOfBoundsException("value cannot be negative");
+      }
+
+      BuilderIndex newEntity = new BuilderIndex(name, value);
+
+      storage.save(newEntity);
+    } else if (value <= associatedEntity.getIndex()) {
+      throw new IndexOutOfBoundsException("value cannot be <= than the previous known value");
```
MForster, line 221: Usually it's better to do all error checking upfront.

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
```diff
@@ -160,4 +162,63 @@ public BuildInfo getRevisionEntry(String commitHash) throws IllegalArgumentExcep
 
     return storage.findById(commitHash, BuildInfo.class);
   }
+
+  /**
+   * {@inheritDoc}
+   */
```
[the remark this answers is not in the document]
MagoDelBlocco: Removed the ones added with these changes, and will remove the rest in a separate PR. Thank you for specifying this!

src/main/java/com/google/graphgeckos/dashboard/storage/DatastoreRepository.java
```diff
@@ -160,4 +162,63 @@ public BuildInfo getRevisionEntry(String commitHash) throws IllegalArgumentExcep
...
+  public int getBuildbotIndex(String buildbotName) throws IllegalArgumentException, NotBoundException {
+    if (buildbotName == null) {
+      throw new IllegalArgumentException("buildbotName cannot be null");
+    }
```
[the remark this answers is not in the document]
MagoDelBlocco: Thank you for pointing out to this library! It's actually really useful for the readability of this component.

Somae (approved):

PR #126, opened 2020-09-09
Rename Prop to Match Expected Name

samuelsogbesan:
This PR rectifies a mismatch between the prop through which the build data is passed to the BuildSnapshot component (<code>data</code>) and prop from which it expected the data (<code>buildData</code>).

This will fix undefined errors resulting from the mismatch.


holounic (approved):
LGTM!

Somae (approved):

PR #132, opened 2020-09-09
Style Header Component

### LINKED ISSUE #30: Style Navigation Component (Header)
Style the Header Component to match the design document.

samuelsogbesan:
This PR integrates the styling for the Header component.

The outcome _(black was used here for illustration, but a CSS variable holding the colour white will be substituted in a later PR_)

_The navigation options here are purely illustrative, but will be merged via #125_
image

Expected result:

image

Closes #30


src/main/webapp/style.css
```diff
@@ -0,0 +1,31 @@
...
+
+.navigation {
+  margin-left:auto;
+}
+
+.navigation > li {
+  padding: 0 2rem;
+  font-style: italic;
+  text-decoration: underline;
+}
```
holounic, line 31: nit: please add newline

holounic (approved):
LGTM

src/main/webapp/style.css
samuelsogbesan: Added with commit b42d472

src/main/webapp/style.css
```diff
@@ -0,0 +1,31 @@
+a {
+  color:inherit;
```
Somae, line 2:
```suggestion
  color: inherit;
```

Somae (approved):
You can consider adding short comments that illustrate what the result / effect of the various CSS classes is summary (not every attribute on it's own, but the combined effect of applying it).

PR #137, opened 2020-09-10
Style Body and Add Colours for Theme

samuelsogbesan:
This PR integrates the basic styling required for the <code>&lt;body></code>.

- **Sets CSS Variables for theme colouring**
- **Set background colour**
- **Sets base text colour**

Closes #136

PR #138, opened 2020-09-10
Change Target Element to Body

samuelsogbesan:
This PR changes the target element which React renders our component to from a <code>div</code> to the main <code>body</code> element.
This changes was made to simplify our DOM structure.
Prior to the change, we would have a structure of **html > body > div > content**
With this change, this is simplified to **html > body > content**

This also means the body will have to be styled accordingly to be treated as a container. See PR #137.

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -4,8 +4,7 @@
     <link rel="stylesheet" href="./style.css">
   </head>
 
-  <body>
-    <div id="root"></div>
+  <body id="root">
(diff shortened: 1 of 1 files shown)


samuelsogbesan:
This behaviour goes against recommended practises for React.
[The body is often modified by external scripts](https://medium.com/@dan_abramov/two-weird-tricks-that-fix-react-7cf9bbdef375), perhaps forcing reloads/re-renders when our application isn't expecting it.

PR #139, opened 2020-09-10
Add missing case to getField

samuelsogbesan:
This PR integrates a missing check in the getField method. 
Before, the <code>obj</code> param wasn't checked to be undefined, and so it was possible to get an error when trying to check obj[field] if obj === undefined.
The fix here is to first check if obj === undefined.

### DIFF
--- a/src/main/webapp/components/utils/getField.js
+++ b/src/main/webapp/components/utils/getField.js
@@ -7,7 +7,9 @@
  * @param defaultValue the value to be used if obj[field] === undefined.
  */
 export const getField = (obj, field, defaultValue) => {
-  return obj[field] !== undefined ?
+  return obj !== undefined ? 
+      obj[field] !== undefined ?
       obj[field] :
-      (defaultValue !== undefined ? defaultValue : null);
+      (defaultValue !== undefined ? defaultValue : null) :
+      defaultValue;
 }
--- a/src/main/webapp/components/utils/getField.js
+++ b/src/main/webapp/components/utils/getField.js
@@ -10,6 +10,6 @@ export const getField = (obj, field, defaultValue) => {
   return obj !== undefined ? 
       obj[field] !== undefined ?
       obj[field] :
-      (defaultValue !== undefined ? defaultValue : null) :
-      defaultValue;
+          (defaultValue !== undefined ? defaultValue : null) :
+          defaultValue;
 }
--- a/src/main/webapp/components/utils/getField.js
+++ b/src/main/webapp/components/utils/getField.js
@@ -7,9 +7,9 @@
  * @param defaultValue the value to be used if obj[field] === undefined.
  */
 export const getField = (obj, field, defaultValue) => {
-  return obj !== undefined ? 
-      obj[field] !== undefined ?
-      obj[field] :
-          (defaultValue !== undefined ? defaultValue : null) :
-          defaultValue;
+  return obj !== undefined
+             ? obj[field] !== undefined
+                   ? obj[field]
+                   : (defaultValue !== undefined ? defaultValue : null)
+             : defaultValue;
 }


src/main/webapp/components/utils/getField.js
```diff
@@ -7,7 +7,9 @@
  * @param defaultValue the value to be used if obj[field] === undefined.
  */
 export const getField = (obj, field, defaultValue) => {
-  return obj[field] !== undefined ?
+  return obj !== undefined ? 
+      obj[field] !== undefined ?
       obj[field] :
-      (defaultValue !== undefined ? defaultValue : null);
+      (defaultValue !== undefined ? defaultValue : null) :
```
MagoDelBlocco, line 13: nit: I'd indent this nested ternary operator more than the one it's based in, for readability purposes

MagoDelBlocco (approved):
LGTM, left a nit

src/main/webapp/components/utils/getField.js
samuelsogbesan: Indented here 8ddf581

Somae (approved):
I can't really comment on how to do this best in JavaScript and will leave it to Martin to comment.

samuelsogbesan:
> I can't really comment on how to do this best in JavaScript and will leave it to Martin to comment.

Should I wait before merging?

src/main/webapp/components/utils/getField.js
```diff
@@ -7,7 +7,9 @@
  * @param defaultValue the value to be used if obj[field] === undefined.
  */
 export const getField = (obj, field, defaultValue) => {
-  return obj[field] !== undefined ?
-      obj[field] :
-      (defaultValue !== undefined ? defaultValue : null);
+  return obj !== undefined
```
mprobst, line 10: the idiomatic pattern is using optional chaining:

```
return obj?.[field] ?? defaultValue;
```

Note that this requires a fairly recent browser.

src/main/webapp/components/utils/getField.js
```diff
@@ -7,7 +7,9 @@
  * @param defaultValue the value to be used if obj[field] === undefined.
```
mprobst, line 7: not related to this CL, but generally speaking accessing properties off of objects using `[]` syntax is an anti-pattern.
* If your code needs to map keys to values, consider using a proper ES6 `Map`.
* If your properties are well known in advance, it's usually not a good idea to attempt to reflect over them: prefer accessing them using regular `.foo`.
* If your properties are potentially `null` or `undefined`
  * the first thing to do is avoid having the problem in the first place. Make sure that all values are set once a value enters the application (e.g. as you parse JSON), and later on always crash the application if a value that should be there isn't.
  * if your application has a value that can (by app semantics) truly not exist, then using `null` or `undefined` is fine. In that case, you can use `??` or `?.` to access it in a reasonably convenient fashion.
  * do not use `?.` syntax to allow null values to flow through your application. It's relatively easy to debug a null crash at the first place where a value appears. It's very hard to debug such a crash several `?.` accesses later.

HTH.

src/main/webapp/components/utils/getField.js
mprobst: ... alternatively, I'd avoid ternaries for better readability and just add more `if` statements:

```
if (!obj) return defaultValue;
if (obj[field] == null) return defaultValue;
...
```

The precedence and association of the cases is much more readable than a ternary in this mode.

samuelsogbesan: Thanks for the insight, learnt something about combining optional chaining with the nullish coalescing operator 😃. 

Based on your other comment, this method might be deprecated in the future. If not, I'll most likely default to a series of <code>if</code> statements since it is far more readable.

src/main/webapp/components/utils/getField.js
samuelsogbesan: Very helpful! I'm considering deprecating <code>getField</code> and <code>getFields</code>. 

Indeed, our app is built so that such values cannot be null / undefined from the way our API serves the data, so perhaps using these methods in this way is redundant.

The original utility of <code>getField</code> was to be used in conjunction with <code>getFields</code> to return a subset of fields from an object. That being said, this can be achieved with **destructing** instead.

> the first thing to do is avoid having the problem in the first place. Make sure that all values are set once a value enters the application (e.g. as you parse JSON), and later on always crash the application if a value that should be there isn't.

Yes I see how letting the app crash in these contexts is very useful for debugging!

Thanks for the insight.

samuelsogbesan:
Please see #176 as to why this PR is closed.

samuelsogbesan:
Please see #176 as to why this PR is closed.

PR #140, opened 2020-09-10
Use getField to Fetch Props

samuelsogbesan:
This PR makes use of the getField utility to safely fetch fields from the props object passed to React components.
This change is beneficial because it prevents the possibility of an unsuspecting 'undefined' error when fetching nested attributes.


holounic (approved):
LGTM

PR #142, opened 2020-09-10
Style Snapshot Component (Head) 

samuelsogbesan:
This PR integrates the core styling requried for the Snapshot Head Component.

Outcome: 

image

_Note the failure group component is yet to be developed, so "Group" is our placeholder._

Excepted Result:

image


holounic (approved):
LGTM!

btw it looks very cool

src/main/webapp/style.css
```diff
@@ -0,0 +1,56 @@
+:root {
+  --failed: #DC4F4F;
```
krasimir-google, line 2: please use lowercase for the constant: https://google.github.io/styleguide/htmlcssguide.html#Capitalization

krasimir-google (approved):
LGTM with a couple of nits

src/main/webapp/style.css
samuelsogbesan: Addressed here d029e00

PR #144, opened 2020-09-10
Fix test failure for deleteRevisionEntry

MagoDelBlocco:
Changed the return value, in case there is no revision with that certain commit hash to delete, to true. The reasoning is that, in the sense of having the end result accomplished (to have the entry with that commit hash no longer in the datastore), deleting the entry, or not having the entry at all, are virtually the same.


samuelsogbesan (approved):
LGTM

MagoDelBlocco:
Solved in #147

PR #145, opened 2020-09-10
Add equals method to BuildBotData

MagoDelBlocco:
Since BuildInfo's equals method depends on the equals method in BuildBotData, DatastoreRepository couldn't be tested properly, making the updateRevisionEntry test fail. This PR adds a concise `equals` method for BuildBotData, fixing the test failure.

The method doesn't check for equality between the commit hash fields, since it's marked as transient for the datastore, as such, there is no guarantee that there is valid data inside of it when retrieving the object.

### DIFF
--- a/src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
@@ -121,6 +121,9 @@ public class BuildBotData {
     this.timestamp = timestamp;
   }
 
+  /**
+   * Checks if all non-transient fields are equal between BuildBotData instances.
+   */
   @Override
   public boolean equals(Object o) {
     if (o == null || !(o instanceof BuildBotData)) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -121,4 +121,18 @@ public void setTimestamp(@NonNull Timestamp timestamp) {
     this.timestamp = timestamp;
   }
 
+  /**
+   * Checks if all non-transient fields are equal between BuildBotData instances.
+   */
+  @Override
+  public boolean equals(Object o) {
```
holounic, line 128: Consider adding ```if (o == this) return true```

holounic (changes requested):

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
MagoDelBlocco: Why? This is unlikely to happen, and if this happens it's covered by the last if. The value we get from adding this special case it's overshadowed by what performance we lose every time we check for this.

holounic: Ok!

holounic (approved):
LGTM!

krasimir-google (approved):
LGTM

PR #146, opened 2020-09-11
Replace ```@NonNull``` annotations with Objects.requireNonNull(o)

holounic:
Replace ```@NonNull``` annotations with Objects.requireNonNull(o) because Spring ```@NonNull``` is an annotation for IDE to prevent explicit null values from being passed to as the annotated args. The ```testRequestsNullData()``` works fine now.


MagoDelBlocco (approved):
LGTM

krasimir-google (commented):
LGTM

Thank you!

krasimir-google (approved):
LGTM

PR #150, opened 2020-09-14
Fix master compilation failure

MagoDelBlocco:
Changed logs from final to mutable, since it has to have a setter


holounic (approved):
LGTM

krasimir-google (approved):
LGTM

PR #151, opened 2020-09-14
Remove final modifier of non-final variable log

holounic:
Fixed the compilation failure described above.

Final fields don't have setters.


holounic:
Closed because this PR duplicates already existing one.

PR #127, opened 2020-09-09
Change endpoint

holounic:
Changed endpoint and ```@PathVariable``` name to make endpoint more descriptive.

PR #157, opened 2020-09-14
Catch GET Request errors and Change Target Endpoint

samuelsogbesan:
This PR adds a .catch at the end of the promise change in case no valid JSON in received.
This PR also changes the target endpoint from /data to /builders, since this is where we send the data from.

See #127 for endpoint information.

### DIFF
--- a/src/main/webapp/components/BuildSnapshotContainer.js
+++ b/src/main/webapp/components/BuildSnapshotContainer.js
@@ -19,7 +19,7 @@ export const BuildSnapshotContainer = React.memo((props) =>
      * @see <a href="www.robinwieruch.de/react-hooks-fetch-data">Fetching</a>
      */
     React.useEffect(() => {
-      fetch(SOURCE).then(res => setData(res.json()));
+      fetch(SOURCE).then(res => setData(res.json())).catch(setData([]));
     }, []);
 
     return (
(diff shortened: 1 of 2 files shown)


holounic (approved):
LGTM

src/main/webapp/components/BuildSnapshotContainer.js
```diff
@@ -19,7 +19,7 @@ export const BuildSnapshotContainer = React.memo((props) =>
      * @see <a href="www.robinwieruch.de/react-hooks-fetch-data">Fetching</a>
      */
     React.useEffect(() => {
-      fetch(SOURCE).then(res => setData(res.json()));
+      fetch(SOURCE).then(res => setData(res.json())).catch(setData([]));
```
krasimir-google, line 22: (suggestion for an improvement in a future PR): consider surfacing some sort of a high-level message that there was a problem fetching data instead of just setting an empty list

krasimir-google (approved):
LGTM

I left a suggestion for a possible improvement in a future PR

src/main/webapp/components/BuildSnapshotContainer.js
samuelsogbesan: +1, this is a good idea, perhaps some feedback on the UI would be applicable.

PR #86, opened 2020-09-03
Replace References to Log Entity with New BuildStep Entity

samuelsogbesan:
This PR integrates a new Entity that represents the step-by-step information for a given build. 
This change comes as a result of a decision to **no longer source the build logs from the top-level "logs" field**, but instead from **the "log" field attached to each child of the steps field**.

#### Justifications for change
- The steps field provides several other pieces of supporting meta-data on top of the logs.
- The logs field doesn't give enough detail to the purpose or meaning of the logs.
- There is not a one-to-one relationship with steps and logs. The steps field helps us identify which logs belong to which step.

Points of contention:
- What should the name of the Entity via the @Entity annotation?

PR #158, opened 2020-09-14
Add equals method to Log

MagoDelBlocco:
Since BuildInfo::equals calls BuildBotData::equals which calls Log::equals is needed in the DatastoreRepository tests, and Log::equals is missing, this caused one test to fail. This PR adds the equals method to the Log class, and fixes the test failure.


samuelsogbesan (approved):
LGTM. The change is needed but its application will possible get superseded by #86 since we are using <code>BuildStep</code> in the place of <code>Log</code>. Outside the scope of this PR, but the addition will have to made for BuildStep.

krasimir-google (approved):
LGTM

PR #160, opened 2020-09-15
Refactor Navigation Structure & Styling

samuelsogbesan:
This PR removes the redundant wrappers from the Navigation Header component and sets the navigation to flex.

### DIFF
--- a/src/main/webapp/style.css
+++ b/src/main/webapp/style.css
@@ -27,7 +27,7 @@ ul {
 
 .navigation {
   display: flex;
-  margin-left: auto;
+  margin: 0 0 0 auto;
 }
 
 .navigation > li {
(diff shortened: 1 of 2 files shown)


src/main/webapp/style.css
```diff
@@ -26,6 +26,7 @@ ul {
 }
 
 .navigation {
+  display: flex;
   margin-left: auto;
```
MagoDelBlocco, line 30: nit (and not the scope of this PR): use the shorthand property margin

MagoDelBlocco (approved):
LGTM, left a nit that is not in the scope of this PR

src/main/webapp/style.css
samuelsogbesan: I think it can be adjusted here since changes were made to the navigation's styling.

samuelsogbesan: Replaced here; Scope of PR adjust accordingly 64a7f58

Somae (approved):

PR #161, opened 2020-09-15
Remove unintended comparisons in equals method

MagoDelBlocco:
BuildInfo::equals was comparing fields annotated as `@Unindexed`, which led to unspecified behaviour. Removed the comparisons with those fields.


samuelsogbesan (approved):
LGTM, Much less comparison!

krasimir-google (approved):
LGTM

PR #162, opened 2020-09-16
Add Missing Parameters to Fetch

### LINKED ISSUE #153: Connect frontend to endpoint

samuelsogbesan:
This PR adds the parameters used to specify the data fetched from <code>/builders</code>

- **NUMBER indicates how many we should take at a time**
- **OFFSET indicates the starting point from which entries are selected relative to the starting index.**

Closes #153

### DIFF
--- a/src/main/webapp/components/BuildSnapshotContainer.js
+++ b/src/main/webapp/components/BuildSnapshotContainer.js
@@ -9,7 +9,11 @@ import {BuildSnapshot} from "./BuildSnapshot";
  */
 export const BuildSnapshotContainer = React.memo((props) =>
   {
-    const SOURCE = '/builders';
+    const NUMBER = 2;
+    const OFFSET = 3;
+
+    const SOURCE = `/builders/number=${NUMBER}/offset=${OFFSET}`;
+    
     const [data, setData] = React.useState([]);
 
     /**
--- a/src/main/webapp/components/BuildSnapshotContainer.js
+++ b/src/main/webapp/components/BuildSnapshotContainer.js
@@ -14,7 +14,7 @@ export const BuildSnapshotContainer = React.memo((props) =>
 
     // The starting point from which the entries are selected.
     // Relative to the starting index.
-    const OFFSET = 3;
+    const OFFSET = 0;
 
     const SOURCE = `/builders/number=${NUMBER}/offset=${OFFSET}`;
 
--- a/src/main/webapp/components/BuildSnapshotContainer.js
+++ b/src/main/webapp/components/BuildSnapshotContainer.js
@@ -10,10 +10,12 @@ import {BuildSnapshot} from "./BuildSnapshot";
 export const BuildSnapshotContainer = React.memo((props) =>
   {
     // Number of entries to be pulled.
+    // TODO: Determine true value for number of entries.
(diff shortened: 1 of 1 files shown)


MagoDelBlocco (approved):
LGTM

src/main/webapp/components/BuildSnapshotContainer.js
```diff
@@ -9,7 +9,15 @@ import {BuildSnapshot} from "./BuildSnapshot";
  */
 export const BuildSnapshotContainer = React.memo((props) =>
   {
-    const SOURCE = '/builders';
+    // Number of entries to be pulled.
+    const NUMBER = 2;
```
krasimir-google, line 13: why is NUMBER 2? I'd expect something like a dozen or more entries on each page?
Could you clarify if this is an intermediate PR just to get the initial integration working and subject to change later?

src/main/webapp/components/BuildSnapshotContainer.js
```diff
@@ -9,7 +9,15 @@ import {BuildSnapshot} from "./BuildSnapshot";
...
+
+    // The starting point from which the entries are selected.
+    // Relative to the starting index.
+    const OFFSET = 0;
```
krasimir-google, line 17: This makes sense as the initial offset. Just to clarify: when the user goes to the next page, I'd expect the UI to query for the next NUMBER entries starting at offset (NUMBER+1). How would this work?

krasimir-google (changes requested):

src/main/webapp/components/BuildSnapshotContainer.js
samuelsogbesan: 2 here is used to get the initial integration working. It will be subject to change later, following a discussion on an appropriate value to use.

src/main/webapp/components/BuildSnapshotContainer.js
samuelsogbesan: I believe this is outside of this PR since it may require some addition React State logic to achieve.

The logic will probably be as follows:
If the user goes to page 2, the NUMBER param will not change, but OFFSET= will increase to be a multiple of NUMBER (e.g. offset for page two is NUMBER * 1). The formula would be **OFFSET=NUMBER * PAGENUMBER-1**

samuelsogbesan: I'll express OFFSET as NUMBER * 0 for the time being.

src/main/webapp/components/BuildSnapshotContainer.js
krasimir-google: consider adding a TODO here that having constants about this here is not the final code state

krasimir-google (approved):
LGTM after adding a TODO about these constants

src/main/webapp/components/BuildSnapshotContainer.js
krasimir-google: Note that OFFSET right now is a const, so this doesn't allow to dynamically change it AFAIK.
Probably we'll need to turn it into React state later.

samuelsogbesan:
> 
> 
> LGTM after adding a TODO about these constants

TODOs added.

PR #163, opened 2020-09-16
enable @Autowired for DatastoreRepository

MagoDelBlocco:
Added a `@Configuration` class for integration with Spring's `@Autowired` annotation, so that Spring can correctly instance DatastoreRepository autowired objects. Returns a DatastoreRepository instanced using a Datastore, with default Datastore options. It is possible that we might need to add some sort of credentials to the options, but this will be seen when testing the endpoint.


samuelsogbesan (approved):
LGTM, thanks for the clear description!

holounic (approved):

PR #164, opened 2020-09-16
Replace mvn deploy command with gcloud deploy in deploy.sh

samuelsogbesan:
This PR replaces the old mvn deploy command with the manual deployment steps.

- <code>mvn clean package</code> builds the application
- <code>gcloud app deploy</code> deploys the application on gcloud.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,3 +1,4 @@
 #!/bin/sh
 gcloud config set project llvm-build-dashboard &&
-mvn package com.google.cloud.tools:appengine-maven-plugin:deploy
+mvn clean package &&
+gcloud app deploy target/step240-2020-1.0.jar


MagoDelBlocco (approved):
LGTM

deploy.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 gcloud config set project llvm-build-dashboard &&
-mvn package com.google.cloud.tools:appengine-maven-plugin:deploy
+mvn clean package &&
```
krasimir-google, line 3: a suggestion for a slight improvement: you can start the bash script with [`set -e`](https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script) and remove all the trailing `&&`-s. This set option has the effect of exiting as soon as the first command executed returns a non-zero exit status.

krasimir-google (approved):
LGTM I've got a suggestion for this bash script that can be done in a separate PR

deploy.sh
samuelsogbesan: Awesome!

PR #165, opened 2020-09-16
Style BuildSnapshot Container

samuelsogbesan:
This PR integrates the styling for the BuildSnapshot Container.

- **Displays container as a flex column**
- **Add gap to 1rem**

It was decided to set this as an ID as it is intended for there to be only one instance of the buildsnapshot container.


holounic (approved):
LGTM!

Somae (approved):

PR #170, opened 2020-09-16
Add BuildbotNotFoundException class

MagoDelBlocco:
This class is supposed to replace `java.rmi.NotBoundException`'s functionality for #121


src/main/java/com/google/graphgeckos/dashboard/storage/BuildbotNotFoundException.java
```diff
@@ -0,0 +1,25 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.graphgeckos.dashboard.storage;
+
+/**
+ * This exception signals that a request has been made for a "index" entry to the database
```
holounic, line 18:
```suggestion
 * This exception signals that a request has been made for an "index" entry to the database
```

src/main/java/com/google/graphgeckos/dashboard/storage/BuildbotNotFoundException.java
```diff
@@ -0,0 +1,25 @@
...
+ * but there is no entry with that key.
+ */
+public class BuildbotNotFoundException extends IllegalArgumentException {
```
holounic, line 21: What does this argument mean? Please rename it so that it reflects its meaning

holounic (changes requested):

src/main/java/com/google/graphgeckos/dashboard/storage/BuildbotNotFoundException.java
MagoDelBlocco: Name taken directly from the Oracle Reference JavaDoc for the constructor that it uses. https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html

holounic: The constructor will still work if you rename this argument to ```message```, for example

holounic (approved):
LGTM

Somae (approved):
LGTM

PR #171, opened 2020-09-17
Style Builder Grid

samuelsogbesan:
This PR integrates the styling for the Builder Grid. Here we use CSSGrid to set the grid to have strictly 3 columns and to generate the rows on the fly as more elements populate the space.


holounic (approved):
LGTM

Somae (approved):
I would still hope that you add more comments that explain semantically what the different styling classes are achieving... I have asked for this on multiple commits now.

samuelsogbesan:
> I would still hope that you add more comments that explain semantically what the different styling classes are achieving... I have asked for this on multiple commits now.

Yes, I will add these in bulk in an incoming PR.

PR #173, opened 2020-09-17
BuildBot fetcher implementation

holounic:
This PR adds the __BuildBotClient__ that is needed to fetch data from the external BuildBot API and add it to the storage in the view of POJO. 

```BuildBotClient``` class contains ```run``` methods that starts fetching information from the BuildBot with the given name.

```BuildBotPopulation``` class allows to easily create a group of BuildBot fetchers with the common base url via ```populate``` function. The information about every BuildBot is provided in the view of an instance of the ```BuildBotInitializer``` class that contains name, initial build id of a build bot and a request frequency.

For testing, the server mock ```MockWebServer``` is used. This eliminates the possibility that tests will fail due to server errors and allows to specify the server's behaviour and verify that the response JSON is parsed correctly. All JSONs were taken from the BuildBot API.

__+1,306 is only because of two long test json files__ 

Tests cover four different test cases:
- When server response is valid, the client should invoke the ```updateRevisionEntry(arg)``` method of the ```DatastoreRepository``` instance, where the arg is the expected POJO.
- When server responds with an error, the client should not invoke the ```updateRevisionEntry(arg)``` method of the ```DatastoreRepository``` instance.
- When server responds with an empty JSON, the client should not invoke the ```updateRevisionEntry(arg)``` method of the ```DatastoreRepository``` instance.
- When the client have enough time to perform more than one request server responds twice with the valid bodies, the client should invoke the ```updateRevisionEntry(arg)``` method of the ```DatastoreRepository``` instance, where the arg is the expected POJO, at least twice.

### CODE REVIEW

File: src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java [hunk 1]
```diff
@@ -203,7 +203,7 @@ public void testGetLastRevisionsQueryAfterDeletion() throws IOException, Interru
     Assert.assertTrue(storage.createRevisionEntry(getDummyGitData("5", time5)));
 
     Assert.assertTrue(storage.deleteRevisionEntry("3"));
-  
+
```
holounic, line 206:
```suggestion
  
```

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
```diff
@@ -0,0 +1,78 @@
+package com.google.graphgeckos.dashboard.fetchers.buildbot;
```
MagoDelBlocco, line 1: Please add the boilerplate license statement in every new file

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
```diff
@@ -0,0 +1,78 @@
...
+   * {@code baseUrl}/{@code buildBot}/builds/{@code buildId}?as_text=1.
+   * E.g with buildBot = "clang-x86_64-debian-fast" and buildId = 1000,
+   * baseUrl = "http://lab.llvm.org:8011/json/builders" the request url will be
+   * http://lab.llvm.org:8011/json/builders/clang-x86_64-debian-fast/builds/1000?as_text=1 .
+   * Adds valid fetched data in the form of {@link BuildBotData} to the storage and tries
+   * to fetch an entry with the next buildId after waiting for {@code delay} seconds.
+   * If the fetched data is empty (invalid) waits for {@code delay} seconds and tries
+   * to make the same request.
+   *
+   * @param buildBot name of the BuildBot as it is in the API (e.g "clang-x86_64-debian-fast").
```
MagoDelBlocco, line 42: remove full stops from `@param` sections in JavaDocs. They are needed for the description and `@return` sections only.

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
+/** A (external) BuildBot API json data fetcher. */
+public class BuildBotClient {
+
+  /** Provides access to the storage. */
+  @Autowired
+  private DatastoreRepository datastoreRepository;
+
+  /** Base url of the BuildBot API. */
```
MagoDelBlocco, line 23: Consider adding a more in-depth explanation of what the base URL is here. Even though it is explained in the constructor, I believe this should be clear in the field explanation as well.

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClientPopulation.java
```diff
@@ -0,0 +1,40 @@
+package com.google.graphgeckos.dashboard.fetchers.buildbot;
+
+import org.springframework.lang.NonNull;
+import java.util.Arrays;
+import java.util.Objects;
+
+/** BuildBot fetchers creator. Create a population of different fetchers with common baseUrl. */
```
MagoDelBlocco, line 7: This responsibility would have been better delivered using the Factory pattern. Consider changing this.

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
MagoDelBlocco, line 206: This change doesn't belong in the scope of this PR

MagoDelBlocco (changes requested):

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClientPopulation.java
holounic: I don't think the factory pattern would be better than the existing one. I prefer leaving it as it is.

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
holounic: Base url is self-explanatory.

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
holounic: I looked through other pieces if our code and didn't found anything controversial about adding fullstops at the end of the ```@param``` description

src/test/java/com/google/graphgeckos/dashboard/storage/DatastoreRepositoryTests.java
holounic:
```suggestion
  
```

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
holounic: Done

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
MagoDelBlocco: Check the discussion in #23. If you've seen this happen in other parts of our code, please fix those problems as well.

MagoDelBlocco (approved):

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
holounic: Thanks! Done

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -76,6 +80,59 @@ public BuildBotData(String commitHash, String name, List<Log> logs, BuilderStatu
...
+    List<Object> changes = (ArrayList<Object>)sourceStamp.get("changes");
+    Map<String, Object> latestChange = (Map<String, Object>)changes.get(0);
+    timestamp = Timestamp.ofTimeMicroseconds(Long.parseLong(latestChange.get("when").toString()));
+  }
+
+  /**
+   * Defines builder status based on a phrase provided in parsed json.
+   * If something failed, then the buildbot is considered failed {@code FAILED}.
+   * If something is lost, then the buildbot is considered lost {@code LOST}.
+   * Otherwise the buildbot is considered passed {@code PASSED}.
```
krasimir-google, line 105: if `words` contains none of `{"failed", "lost", "successful"}`, the status is not updated. This comment hints that the buildbot is considered `PASSED` in this case

I'd suggest extracting a little helper function that returns the status given the list of `words` similar to the body of this one, that `return`-s where this one `break`-s and that returns `PASSED` by default, when none of the words matched `failed` or `lost`.

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
```diff
@@ -0,0 +1,104 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.graphgeckos.dashboard.fetchers.buildbot;
+
+import com.google.graphgeckos.dashboard.datatypes.BuildBotData;
+import com.fasterxml.jackson.databind.ObjectMapper;
```
krasimir-google, line 18: this import is out of sort order

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
krasimir-google: I think it's worth repeating the base URL part of the example from the constructor here. Specifically, consider this extract from the example:
>baseUrl = "http://lab.llvm.org:8011/json/builders"
Someone not familiar with the LLVM BuildBot URLs can reasonably think they can pass any of the following as baseUrl:
* http://lab.llvm.org:8011
* http://lab.llvm.org:8011/json
* http://lab.llvm.org:8011/json/builders

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
```diff
@@ -0,0 +1,104 @@
...
+      .bodyToMono(String.class)
+      .delaySubscription(Duration.ofSeconds(delay))
+      .onErrorResume(e -> {
+        logger.info("Ignoring error: " + e.getMessage());
+        return Mono.empty();
+      })
+      .repeat()
+      .subscribe(response -> {
+        if (response.isEmpty()) {
+          logger.info(String.format("Error occured, waiting in %d seconds", delay));
```
krasimir-google, line 83:
```suggestion
          logger.info(String.format("Error occured, waiting for %d seconds", delay));
```

src/test/java/com/google/graphgeckos/dashboard/buildbot/BuildBotClientTestInfo.java
```diff
@@ -0,0 +1,90 @@
...
+      );
+    }
+    commitHash = expected[0];
+    timestamp = Timestamp.ofTimeMicroseconds(Long.parseLong(expected[1]));
+    name = expected[2];
+    status = BuilderStatus.valueOf(expected[3]);
+    logs = new ArrayList<>();
+
+    for (int i = 4; i < expected.length; i += 2) {
+      logs.add(new Log(expected[i], expected[i + 1]));
```
krasimir-google, line 63: If `expected.length` is odd and larger than 4 this will fail. Consider detecting this when checking the validity of `expected` and throw an IllegalArgumentException in this case

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
```diff
@@ -0,0 +1,104 @@
...
+   * to make the same request.
+   *
+   * @param buildBot name of the BuildBot as it is in the API (e.g "clang-x86_64-debian-fast")
+   * @param initialBuildId the id of the BuildBot's build from where to start fetching data
+   */
+  public void run(@NonNull String buildBot, long initialBuildId, long delay) {
+    Objects.requireNonNull(buildBot);
+
+    AtomicLong buildId = new AtomicLong(initialBuildId);
+    logger.info("Started fetching");
```
krasimir-google, line 66: I assume this `run` method will be called for different instances concurrently.
To make the logging more useful, consider logging some information to help distinguish between those, such as the `baseUrl`, `buildBot` and the `buildId` that you try fetching.

pom.xml
```diff
@@ -113,6 +113,26 @@
       <version>5.6.2</version>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-webflux</artifactId>
```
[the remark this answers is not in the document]
holounic: The WebClient is a part of it

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
```diff
@@ -0,0 +1,104 @@
...
+   * If the fetched data is empty (invalid) waits for {@code delay} seconds and tries
+   * to make the same request.
+   *
+   * @param buildBot name of the BuildBot as it is in the API (e.g "clang-x86_64-debian-fast")
+   * @param initialBuildId the id of the BuildBot's build from where to start fetching data
+   */
+  public void run(@NonNull String buildBot, long initialBuildId, long delay) {
+    Objects.requireNonNull(buildBot);
+
+    AtomicLong buildId = new AtomicLong(initialBuildId);
```
krasimir-google, line 65: I believe this doesn't have to be atomic. It looks like it's potentially incremented in a separate thread, but the pipeline below doesn't look like it's going to be branching and scheduling multiple concurrent executions of the `subscribe` callback, hence there is no risk of a data race on the value of this variable. Note that I may be missing something here, so if you have concrete reasone where this should necessarily be atomic, please share.

src/test/java/com/google/graphgeckos/dashboard/buildbot/BuildBotClientTest.java
```diff
@@ -0,0 +1,266 @@
...
+     * Expected output fields. See {@link com.google.graphgeckos.dashboard.datatypes.BuildBotData} to learn more.
+     */
+    private String commitHash;
+    private Timestamp timestamp;
+    private String name;
+    private BuilderStatus status;
+    private List<Log> logs;
+
+    public BuildBotDataMatcher(BuildBotClientTestInfo expected) {
+      super();
```
krasimir-google, line 165: I believe you don't need to call the default `super` constructor; this is done automatically.

pom.xml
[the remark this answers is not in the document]
krasimir-google: Thank you!

src/test/java/com/google/graphgeckos/dashboard/buildbot/BuildBotClientTest.java
holounic: Removed, thanks!

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
holounic: It has to be atomic because in Java variable used in lambda expression should be final or effectively final (which means being atomic or final).

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
holounic: Done, thanks!

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
holounic: Fixed

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
holounic: Done, thanks!

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -76,6 +80,59 @@ public BuildBotData(String commitHash, String name, List<Log> logs, BuilderStatu
...
   }
 
+  public BuildBotData(@JsonProperty("builderName") String name) {
+    this.name = name;
+  }
+
+  /**
+   * Extracts nested commitHash("revision"), branch("branch") and
+   * timestamp("when")fields from the json@param sourceStamp.
+   * Representation of the json component, where the commitHash field is located.
```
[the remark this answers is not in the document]
holounic: Fixed!

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -76,6 +80,59 @@ public BuildBotData(String commitHash, String name, List<Log> logs, BuilderStatu
...
+    Map<String, Object> latestChange = (Map<String, Object>)changes.get(0);
+    timestamp = Timestamp.ofTimeMicroseconds(Long.parseLong(latestChange.get("when").toString()));
+  }
+
+  /**
+   * Defines builder status based on a phrase provided in parsed json.
+   * If something failed, then the buildbot is considered failed {@code FAILED}.
+   * If something is lost, then the buildbot is considered lost {@code LOST}.
+   * Otherwise the buildbot is considered passed {@code PASSED}.
+   * @param words Words of the "text" JSON field.
```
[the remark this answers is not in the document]
holounic: Done

src/test/java/com/google/graphgeckos/dashboard/buildbot/BuildBotClientTestInfo.java
holounic: Done

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
krasimir-google: I don't see any changes to this comment nor to the `extractStatus` method below. Maybe the PR was not refreshed?

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
krasimir-google: Thank you!

krasimir-google (changes requested):

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -76,6 +80,63 @@ public BuildBotData(String commitHash, String name, List<Log> logs, BuilderStatu
...
+    Map<String, Object> latestChange = (Map<String, Object>)changes.get(0);
+    timestamp = Timestamp.ofTimeMicroseconds(Long.parseLong(latestChange.get("when").toString()));
+  }
+
+  /**
+   * Defines builder status based on a phrase provided in parsed json.
+   * If something failed or the phrase doesn't contain any of the key words
+   * ("failed", "successful", "lost"), then the buildbot is considered failed {@code FAILED}.
+   * If something is lost, then the buildbot is considered lost {@code LOST}.
+   * If  the buildbot is considered passed {@code PASSED}.
```
krasimir-google, line 107: I think there is a part of this sentence missing.

krasimir-google (commented):
Looks good with a comment nit, also in the PR description you could remove the paragraph about `BuildBotPopulation` as that doesn't appear here.

Please wait for a readability review from MForster.

krasimir-google:
LGTM

krasimir-google (approved):

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -76,6 +80,63 @@ public BuildBotData(String commitHash, String name, List<Log> logs, BuilderStatu
...
+   *
+   * @param sourceStamp Representation of the json component, where the commitHash field is located
+   */
+  @JsonProperty("sourceStamp")
+  @SuppressWarnings("unchecked")
+  public void unpackSourceStamp(Map<String, Object> sourceStamp) {
+    commitHash = sourceStamp.get("revision").toString();
+    List<Object> changes = (ArrayList<Object>)sourceStamp.get("changes");
+    Map<String, Object> latestChange = (Map<String, Object>)changes.get(0);
+    timestamp = Timestamp.ofTimeMicroseconds(Long.parseLong(latestChange.get("when").toString()));
```
MForster, line 99: Same here.

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -76,6 +80,63 @@ public BuildBotData(String commitHash, String name, List<Log> logs, BuilderStatu
...
+   * If something failed or the phrase doesn't contain any of the key words
+   * ("failed", "successful", "lost"), then the buildbot is considered failed {@code FAILED}.
+   * If something is lost, then the buildbot is considered lost {@code LOST}.
+   * If  the buildbot is considered passed {@code PASSED}.
+   *
+   * @param words Words of the "text" JSON field
+   */
+  @JsonProperty("text")
+  private void extractStatus(List<String> words) {
+    for (String word : words) {
```
MForster, line 113: Is it possible that the list is empty?

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -76,6 +80,63 @@ public BuildBotData(String commitHash, String name, List<Log> logs, BuilderStatu
...
+
+  /**
+   * Unpacks logs represented as list of lists of two strings,
+   * where the first one is a type of the log and the second one
+   * is a link to the log.
+   *
+   * @param logs Representation of the json component, where the logs are located
+   */
+  @JsonProperty("logs")
+  private void unpackLogs(List<String[]> logs) {
```
MForster, line 136: A list of arrays is an usual data structure. Do you get that from some library?

src/main/java/com/google/graphgeckos/dashboard/fetchers/buildbot/BuildBotClient.java
```diff
@@ -0,0 +1,106 @@
...
+  @Autowired
+  private DatastoreRepository datastoreRepository;
+
+  /** Base url of the BuildBot API, LLVM BuildBot API base url is "http://lab.llvm.org:8011/json/builders" */
+  private String baseUrl;
+
+  private static final Logger logger = Logger.getLogger(BuildBotClient.class.getName());
+
+  public BuildBotClient(@NonNull String baseUrl) {
+    Objects.requireNonNull(baseUrl);
```
MForster, line 44: `Preconditions.checkNotNull()` is preferred

MForster: and then use it directly in the assignment.

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
holounic: I haven't seen that in real BuildBot API JSONs, but in this case the BuildBot will be considered as failed.

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -76,6 +80,63 @@ public BuildBotData(String commitHash, String name, List<Log> logs, BuilderStatu
...
+  /**
+   * Extracts nested commitHash("revision"), branch("branch") and
+   * timestamp("when") fields from the json
+   *
+   * @param sourceStamp Representation of the json component, where the commitHash field is located
+   */
+  @JsonProperty("sourceStamp")
+  @SuppressWarnings("unchecked")
+  public void unpackSourceStamp(Map<String, Object> sourceStamp) {
+    commitHash = sourceStamp.get("revision").toString();
```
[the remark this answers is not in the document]
holounic: Fixed, thank you!

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
holounic: No, I thought that the String[] is an Object so it could be used as a type of a generic. But I still replaced it by the List<List<String>>

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildBotData.java
```diff
@@ -76,6 +80,63 @@ public BuildBotData(String commitHash, String name, List<Log> logs, BuilderStatu
...
+  }
+
+  /**
+   * Extracts nested commitHash("revision"), branch("branch") and
+   * timestamp("when") fields from the json
+   *
+   * @param sourceStamp Representation of the json component, where the commitHash field is located
+   */
+  @JsonProperty("sourceStamp")
+  @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
holounic: Fixed

PR #175, opened 2020-09-17
Change BuildSnapshot Styling with onclick event

samuelsogbesan:
This PR adds a change that adjusts the orientation of the toggle icon and the composition of the borders for the BuildSnapshot Component. This is all bound together using a className that is conditioned to appear only when <code>props.isOpen === true</code>
- Smoothly rotate the toggle icon on click
- Reveal snapshot header border bottom on click.


src/main/webapp/style.css
```diff
@@ -70,6 +70,15 @@ ul {
...
   width: 1.5rem;
+  transition: 300ms transform ease-in-out;
+}
+
+.snapshot-header.active > .header-toggle {
+  transform: rotate(90deg);
+}
+
+.snapshot-header.active {
+  border-bottom: 0.25rem var(--primary) solid;
```
MagoDelBlocco, line 81: nit: use .25 instead of 0.25

MagoDelBlocco (approved):
LGTM, left a nit

src/main/webapp/style.css
samuelsogbesan: Adjust in all places! 46f2630

src/main/webapp/style.css
```diff
@@ -70,6 +70,16 @@ ul {
 .header-toggle {
   height: 1.5rem;
   width: 1.5rem;
+  transition: 300ms transform ease-in-out;
+}
+
+/*Animate the open/close icon when snapshot is open*/
```
Somae, line 76:
```suggestion
/* Animate the open/close icon when snapshot is open */
```
See https://google.github.io/styleguide/htmlcssguide.html#Protocol for CSS comment examples with surrounding whitespaces.

src/main/webapp/style.css
```diff
@@ -51,14 +51,14 @@ ul {
 
 .snapshot-header {
   align-items: center;
-  border: 0.25rem var(--primary) solid;
-  box-shadow: 0 0.25rem 0.25rem rgba(0, 0, 0, 0.25);
-  border-radius: 0.5rem;
+  border: .25rem var(--primary) solid;
+  box-shadow: 0 .25rem .25rem rgba(0, 0, 0, .25);
+  border-radius: .5rem;
```
Somae, line 56: Should this be in alphabetical order?

src/main/webapp/style.css
Somae: Please also add what kind of animation it is. Animate the [] by [] for example.

Somae (changes requested):

src/main/webapp/style.css
samuelsogbesan: I've updated the comment to be more specific d8a307d

src/main/webapp/style.css
samuelsogbesan: Silly mistake! d8a307d

Somae (approved):

PR #176, opened 2020-09-18
Deprecate getField & getFields

samuelsogbesan:
This PR introduces a major change in how data is sourced between components and our API.

### Motivation 

- It is preferable for our component's props to naturally be type safe by santising input as data enters the application rather than catching every potentially case of <code>null</code> and <code>undefined</code> using <code>getField</code> and <code>getFields</code>. 

- The shape of our data makes it impossible to receive <code>null</code> values from the API.

- From a debugging perspective, letting type errors bubble through will give us an indication of a poorly formed piece of data, which otherwise would've been obscured by <code>defaultValue</code>.

### Solution

- Deprecate <code>getField</code> and  <code>getFields</code>.

- Replace them with Javascript's built in object [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) and [property accessors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors).

I highly recommend reading the comments by @mprobst on PR #139 about [optional chaining](https://github.com/googleinterns/step240-2020/pull/139?w=1#discussion_r488569738), [ternaries & readability](https://github.com/googleinterns/step240-2020/pull/139?w=1#discussion_r488573798) and [property accessor anti-patterns & dealing with unexpected values](https://github.com/googleinterns/step240-2020/pull/139?w=1#discussion_r488572692) for more context as to why the change was necessary.


src/main/webapp/components/utils/getFields.js
```diff
@@ -1,13 +1,15 @@
...
-  * @param fields an array of the target fields, in string form.
-  * @param defaultValue the value to be used if obj[field] === undefined.
+ * Creates a new object whose fields are either equal to obj[field]
+ * or defaultValue.
+ * Useful for extracting a subset of fields from an object.
+ *
+ * @param obj the object to query the fields for.
+ * @param fields an array of the target fields, in string form.
+ * @param defaultValue the value to be used if obj[field] === undefined.
+ * @deprecated extracting values with destructuring instead.
```
mprobst, line 11: nit: typo, should be 'extract'

mprobst (approved):

src/main/webapp/components/utils/getFields.js
samuelsogbesan: Spelling adjusted with commit 7fa0822.

holounic (approved):
LGTM

krasimir-google (approved):
Cool!

PR #177, opened 2020-09-18
Add User Feedback for Results of Fetch

samuelsogbesan:
This PR integrates some feedback to the user in the case that no data is received from the API.
Previously, users would get an empty screen. Now, they are provided a prompt telling them no data has been received.

image

### DIFF
--- a/src/main/webapp/components/BuildSnapshotContainer.js
+++ b/src/main/webapp/components/BuildSnapshotContainer.js
@@ -37,7 +37,7 @@ export const BuildSnapshotContainer = React.memo((props) =>
 
     let content;
     if (data.length > 0) {
-      content = data.map(snapshotData => <BuildSnapshot buildData={snapshotData}/>)
+      content = data.map(snapshot => <BuildSnapshot buildData={snapshot}/>);
     } else {
       content = <span className='loader'>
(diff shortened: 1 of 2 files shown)


holounic (approved):
LGTM

src/main/webapp/components/BuildSnapshotContainer.js
```diff
@@ -35,9 +35,17 @@ export const BuildSnapshotContainer = React.memo((props) =>
         .catch(setData([]));
     }, []);
 
+    let content;
+    if (data.length > 0) {
+      content = data.map(snapshot => <BuildSnapshot buildData={snapshot}/>);
+    } else {
+      content = <span className='loader'>
+          No new revisions to display as of {new Date().toString()}.</span>;
```
Somae, line 43: Is this displaying the date in the user's time zone?

Somae (approved):

src/main/webapp/components/BuildSnapshotContainer.js
samuelsogbesan: Yes!

samuelsogbesan: Replaced in this commit 00dabaa

PR #180, opened 2020-09-21
Create main test class to run all tests of the application

holounic:
It's nice to be able to run all the tests at once, because it makes sure that no part of the application has been broken by the changes.


samuelsogbesan:
When you run <code>mvn clean package</code> to build the application, all these tests already run. What is the use case here?

PR #187, opened 2020-09-22
Add Licensing Blurb to files (JS & CSS)

samuelsogbesan:
This PR adds the Apache Licensing notice at the top of our front-end files, including our React Components and Stylesheets.

#### Point of Contention
Does this need to be applied also to <code>package.json</code> and <code>package-lock.json</code>? What about our deprecated units (<code>getField</code> and <code>getFields</code>)?

### DIFF
--- a/src/main/webapp/App.js
+++ b/src/main/webapp/App.js
@@ -1,3 +1,17 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


MagoDelBlocco (approved):

MagoDelBlocco:
Regarding the boilerplate license on `package.json` and `package-lock.json`, they are configuration files, so I don't think we should include the license in those.

Somae (approved):
I agree that our configuration files result from the copyrighted code, but I could try to find out if those should be copyrighted anyway. This PR can go ahead regardless!

PR #188, opened 2020-09-22
Remove props parameter where props are not expected

samuelsogbesan:
This PR removes the <code>props</code> argument from functional components that expect no props. 
This PR also removes a placeholder prop left over from bootstrapping 😄


MagoDelBlocco (approved):
LGTM

krasimir-google (approved):
LGTM

PR #189, opened 2020-09-22
Refactor Destructuring

samuelsogbesan:
This PR changes the destructing method using in BuildSnapshot to destructure <code>props.buildData</code> once and explicitly create the new <code>headerData</code> and <code>trayData</code> objects.

#### Motivation for change
- The previous method seems to not be fully supported. Although it worked in console (chrome dev tools), it threw an error at runtime.
- No longer have to destructure buildData twice.


holounic (approved):
LGTM

krasimir-google (approved):
LGTM

PR #191, opened 2020-09-22
Add Key to Components Generated with Array.map

samuelsogbesan:
To suppress the <code>react-no-key-warning</code>, this PR adds a key wherever <code>Array.map</code> is used to recursively generate React components.

[The React Documentation outlines why keys are necessary, and can have a huge performance boost when used correctly](https://reactjs.org/docs/reconciliation.html#recursing-on-children)


MagoDelBlocco (approved):
LGTM

Somae (approved):

PR #192, opened 2020-09-22
Add pagination functionality

MagoDelBlocco:
This PR adds the structure needed for changing the pages shown on the front end. **This is done via 2 buttons "prevPage" and "nextPage", and a visual indicator of the current page number.**

Currently, each page has a fixed number of **20 build snapshots**, but the plan is for the user to be able to choose how many entries per page it wants. The styling will come in a later PR.

### DIFF
--- a/src/main/webapp/components/BuildSnapshotContainer.js
+++ b/src/main/webapp/components/BuildSnapshotContainer.js
@@ -24,7 +24,7 @@ export const BuildSnapshotContainer = React.memo((props) =>
     const SOURCE = `/builders/number=${NUMBER}/offset=${OFFSET}`;
 
     /**
-     * Moves pageNumber with a set increment
+     * Moves pageNumber with a set increment.
      *
      * @param increment is how many pages to jump over
      *     (negative values for going back)
--- a/src/main/webapp/components/BuildSnapshotContainer.js
+++ b/src/main/webapp/components/BuildSnapshotContainer.js
@@ -13,7 +13,7 @@ export const BuildSnapshotContainer = React.memo((props) =>
     const [data, setData] = React.useState([]);
 
     // Number of entries to be pulled. 20 seems reasonable enough.
-    // TODO: Add options for this number on paginationNavbar
+    // TODO: Add options for this number on paginationNavbar.
     const NUMBER = 20;
 
     // The starting point from which the entries are selected.
--- a/src/main/webapp/components/BuildSnapshotContainer.js
+++ b/src/main/webapp/components/BuildSnapshotContainer.js
@@ -20,7 +20,7 @@ export const BuildSnapshotContainer = React.memo((props) =>
     // Relative to the starting index.
     const OFFSET = NUMBER * pageNumber;
 
-    // The endpoint from where we get data from
+    // The endpoint from where we get data from.
     const SOURCE = `/builders/number=${NUMBER}/offset=${OFFSET}`;
 
     /**
--- a/src/main/webapp/components/BuildSnapshotContainer.js
+++ b/src/main/webapp/components/BuildSnapshotContainer.js
@@ -9,18 +9,27 @@ import {BuildSnapshot} from "./BuildSnapshot";
  */
 export const BuildSnapshotContainer = React.memo((props) =>
   {
-    // Number of entries to be pulled.
-    // TODO: Determine true value for number of entries.
-    const NUMBER = 2;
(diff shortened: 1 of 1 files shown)


src/main/webapp/components/BuildSnapshotContainer.js
```diff
@@ -43,10 +52,27 @@ export const BuildSnapshotContainer = React.memo((props) =>
           No new revisions to display as of {new Date().toLocaleString()}.</span>;
     }
 
+    let hasPrev = pageNumber == 0;
```
samuelsogbesan, line 55: Use <code>===</code> for [equality checks](https://google.github.io/styleguide/jsguide.html#features-equality-checks)
```suggestion
    let hasPrev = pageNumber === 0;
```

samuelsogbesan (approved):
LGTM, bar one nit.

I'd recommend we move the pagination-navbar to render conditionally with the build-snaphot-container. This way if no data is returned, we don't still have the pagination section. Also, it would be better if the pagination tooling came above the build-snapshot-container, otherwise it will get pushed to the bottom and won't be visible when we start getting data in.

This change can be for a future PR because it'll require the logic to shuffle around a little bit. 

i'd also recommend in the future to just add the paginationNavbar directly into the div rather than setting it up externally, for readability.

Good job though :)

MagoDelBlocco:
> I'd recommend we move the pagination-navbar to render conditionally with the build-snaphot-container. This way if no data is returned, we don't still have the pagination section. 

Theoretically, this app should always have data once we properly deploy it, and the only cases where we don't have any data to show would be in test builds and now, in the beginning. I don't think this warrants the addition of new logic just for this case.

MagoDelBlocco:
> I'd recommend we move the pagination-navbar to render conditionally with the build-snaphot-container. This way if no data is returned, we don't still have the pagination section. 

Addressed this by making the buttons invisible instead of grey-ed out, in #193. No additional logic needs to be added.

MagoDelBlocco:
Opened issue #208, which aims to move the Pagination structure to its own component.

src/main/webapp/components/BuildSnapshotContainer.js
```diff
@@ -9,18 +9,27 @@ import {BuildSnapshot} from "./BuildSnapshot";
...
 export const BuildSnapshotContainer = React.memo((props) =>
   {
-    // Number of entries to be pulled.
-    // TODO: Determine true value for number of entries.
-    const NUMBER = 2;
+    const [pageNumber, setPageNumber] = React.useState(0);
+    const [data, setData] = React.useState([]);
+
+    // Number of entries to be pulled. 20 seems reasonable enough.
+    // TODO: Add options for this number on paginationNavbar
```
Somae, line 16:
```suggestion
    // TODO: Add options for this number on paginationNavbar.
```

src/main/webapp/components/BuildSnapshotContainer.js
```diff
@@ -9,18 +9,27 @@ import {BuildSnapshot} from "./BuildSnapshot";
...
-    // TODO: vary multiplier based on page number.
-    const OFFSET = NUMBER * 0;
+    const OFFSET = NUMBER * pageNumber;
 
+    // The endpoint from where we get data from
     const SOURCE = `/builders/number=${NUMBER}/offset=${OFFSET}`;
 
-    const [data, setData] = React.useState([]);
+    /**
+     * Moves pageNumber with a set increment
```
Somae, line 27:
```suggestion
     * Moves pageNumber with a set increment.
```

src/main/webapp/components/BuildSnapshotContainer.js
```diff
@@ -9,18 +9,27 @@ import {BuildSnapshot} from "./BuildSnapshot";
...
+    // TODO: Add options for this number on paginationNavbar
+    const NUMBER = 20;
 
     // The starting point from which the entries are selected.
     // Relative to the starting index.
-    // TODO: vary multiplier based on page number.
-    const OFFSET = NUMBER * 0;
+    const OFFSET = NUMBER * pageNumber;
 
+    // The endpoint from where we get data from
```
Somae, line 23:
```suggestion
    // The endpoint from where we get data from.
```

Somae (approved):

MagoDelBlocco:
This PR got "emergency merged" to provide the functionality needed for the live demo. It was added in the list at #214 for PRs that need readability re-review.

PR #193, opened 2020-09-22
Add styling for pagination

MagoDelBlocco:
Added styling for the pagination buttons, having the next and previous page buttons to the bottom right of the page, with a proper size for ease of use, and kept the page number to the left of the page. Colors for the buttons are greyed-out if the change cannot be made (user already on last/first page), while keeping them white if they are clickable.

Screenshot 2020-09-22 at 19 29 38 - Display 2

### DIFF
--- a/src/main/webapp/style.css
+++ b/src/main/webapp/style.css
@@ -153,9 +153,6 @@ ul {
 /* Pagination Button Element when hidden */
 .hiddenButton {
   background: transparent;
-  border-width: 0 var(--button-size) var(--button-size) 0;
-  float: right;
-  padding: var(--button-size);
 }
 
 /* Pagination Button Element when visible */
--- a/src/main/webapp/style.css
+++ b/src/main/webapp/style.css
@@ -4,10 +4,10 @@
 
 :root {
   --background: #0c0d25;
+  --button-size: .3125rem;
   --failed: #dc4f4f;
   --passed: #4fdc7f;
   --primary: white;
-  --button-size: .3125rem;
   font-size: 16px;
 }
 
--- a/src/main/webapp/style.css
+++ b/src/main/webapp/style.css
@@ -148,3 +148,33 @@ ul {
   padding: 1rem 0;
   text-align: center;
 }
+
+.hiddenButton {
+  border: solid #333;
+  border-width: 0 5px 5px 0;
+  float: right;
+  padding: 5px;
+}
+
+.paginationButton {
+  border: solid white;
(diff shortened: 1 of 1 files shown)


src/main/webapp/style.css
```diff
@@ -148,3 +148,33 @@ ul {
...
+
+.hiddenButton {
+  border: solid #333;
+  border-width: 0 5px 5px 0;
+  float: right;
+  padding: 5px;
+}
+
+.paginationButton {
+  border: solid white;
```
samuelsogbesan, line 160: I recommend we set the border width here. Also instead of using <code>white</code>, you could use the CSS variable <code>primary</code> setup in <code>.root</code>. This way the colour can be adjusted with the rest of the theme.

samuelsogbesan (commented):
Stylistically, this is great I have some comments on the style here.

To hide the button I'd recommend using something like<code>.paginationButton.hidden</code> as the selector and toggling <code>.hidden</code> in the React component. This way the button style is maintained, and all that has to change are the attributes we use to hide the button (which in this case is the border?).

Also, adding some comments would be great.

src/main/webapp/style.css
```diff
@@ -148,3 +148,33 @@ ul {
   padding: 1rem 0;
   text-align: center;
 }
+
+.hiddenButton {
+  border: solid #333;
```
samuelsogbesan, line 153: Specify border size using <code>rem units</code>. Rem units are a ratio of the font size declared in the <code>:root</code> (16px), e.g. 1rem = 16px.

src/main/webapp/style.css
MagoDelBlocco: Done, thank you!

MagoDelBlocco:
Because of the way the arrows are created, centering the arrows would require structural changes. I'll file an issue to change this, but for the demo tomorrow we'll have them on the right side of the page.

Somae (approved):

src/main/webapp/style.css
```diff
@@ -7,6 +7,7 @@
   --failed: #dc4f4f;
   --passed: #4fdc7f;
   --primary: white;
+  --button-size: .3125rem;
```
samuelsogbesan, line 10: Please put this in it's alphabetical place before merging

samuelsogbesan (approved):
LGTM - address nit before merging please

src/main/webapp/style.css
MagoDelBlocco: Done. Thank you!

PR #195, opened 2020-09-23
Remove Unneeded Import in DatastoreRepositoryTests

MagoDelBlocco:
This PR removes `import org.junit.Before` in DatastoreRepositoryTests, since that functionality is not used anymore. In the tests, the emulator is always cleaned up **after** each test.


holounic (approved):
LGTM

krasimir-google (approved):

PR #198, opened 2020-09-23
Fix Header Styling

samuelsogbesan:
This PR removes some left over style rules that broke the styling for the header.


MagoDelBlocco (approved):
LGTM

krasimir-google (approved):

PR #200, opened 2020-09-23
Update Javadocs for DataRepository

MagoDelBlocco:
Updated the `@throws` annotations for the index related methods in DatastoreRepository. Now they accurately reflect what exceptions can be thrown.


samuelsogbesan (approved):
LGTM

PR #202, opened 2020-09-23
Update DatastoreRepository Class JDoc

MagoDelBlocco:
Changed documentation to reflect the new addition of "index" type entry, and removed an easily deprecable paragraph about relevant "revision" fields.


samuelsogbesan (approved):
LGTM, neat changes

PR #204, opened 2020-09-24
Add a config file with initial build bot data

holounic:
Added config file with four buildbots, functionality to read this file and create buildbots with the given name, initial id and delay and the corresponding test.


src/main/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpReader.java
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Creates list of data that is needed for initializing and running instances of
+   * {@link com.google.graphgeckos.dashboard.fetchers.buildbot.BuildBotClient}.
+   *
+   * @return a list of data for initializing BuildBot fetchers.
+   */
+  public static List<BuildBotSetUpData> read() {
+    try {
+      return deserialize(Files.lines(Paths.get(filePath)).collect(Collectors.joining("\n")));
```
Somae, line 39: I only skimmed this PR so far as Sam hasn't looked at the PR yet, but you could try to make sure to allow for cross-platform compatible line endings. I made a quick search and found this: https://intellipaat.com/community/32399/is-there-a-newline-constant-defined-in-java-like-environment-newline-in-c#:~:text=Java%20API%20%3A%20System.lineSeparator&text=It%20always%20returns%20the%20same,%22%5Cr%5Cn%22.

The potential issue with \n is that it assumes a UNIX style system.

src/main/resources/static/buildbots.txt
```diff
@@ -0,0 +1,12 @@
+clang-x86_64-debian-fast
```
Somae, line 1: Please make sure to document the expected input style of this file, also in the corresponding reader.

src/main/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpReader.java
holounic: Done!

Somae: You only replaced the line separator on a different part of the file, but not here where my comment is =).

src/main/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpReader.java
```diff
@@ -0,0 +1,50 @@
...
+  /** Path to the config file with the initial BuildBot data. */
+  private static String filePath = "src/main/resources/static/buildbots.txt";
+
+  private BuildBotSetUpReader() {}
+
+  private static List<BuildBotSetUpData> deserialize(String info) {
+    List<BuildBotSetUpData> bots = new ArrayList<>();
+    String[] splited = info.split(System.lineSeparator());
+    if (splited.length < 3 || splited.length % 3 != 0) {
+      throw new IllegalArgumentException("Every buildbot has a name, an initial id and a delay");
```
Somae, line 22:
```suggestion
      throw new IllegalArgumentException("Every buildbot needs a name, an initial id and a delay on separate lines");
```

Somae: Or something like this.

src/main/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpReader.java
holounic: Ooops, sorry. Fixed it in the latest commit.

src/test/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpReaderTest.java
```diff
@@ -0,0 +1,27 @@
+package com.google.graphgeckos.dashboard.reader;
```
Somae, line 1: It would be great to have a test that verifies the behavior for a malformed input file, but you might skip this given the short remaining time.

src/main/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpReader.java
```diff
@@ -0,0 +1,59 @@
...
+    return bots;
+  }
+
+  /**
+   * Creates list of data that is needed for initializing and running instances of
+   * {@link com.google.graphgeckos.dashboard.fetchers.buildbot.BuildBotClient}.
+   *
+   * @return a list of data for initializing BuildBot fetchers.
+   */
+  public static List<BuildBotSetUpData> read() {
```
MForster, line 46: Avoid static methods. They make testing harder (mocking, for example)

src/main/java/com/google/graphgeckos/dashboard/DashboardApplication.java
```diff
@@ -1,17 +1,29 @@
...
 import org.springframework.cloud.gcp.data.datastore.repository.config.EnableDatastoreRepositories;
 import org.springframework.context.annotation.Configuration;
+import java.util.List;
 
 @SpringBootApplication
 @Configuration
 @EnableDatastoreRepositories
 public class DashboardApplication {
 
+  private static final String BASE_URL = "http://lab.llvm.org:8011/builders";
```
MForster, line 17: This should probably go into command line options or into a config file.

src/main/java/com/google/graphgeckos/dashboard/DashboardApplication.java
```diff
@@ -1,17 +1,29 @@
...
+
   public static void main(String[] args) {
     SpringApplication.run(DashboardApplication.class, args);
+
+    List<BuildBotSetUpData> buildBotToRun = BuildBotSetUpReader.read();
+    if (buildBotToRun == null) {
+      throw new IllegalArgumentException("List of BuildBots to run can't be null. Check the config file.");
```
MForster, line 24: Throwing an exception seems not appropriate here. I'd rather just print an error message.

src/main/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpData.java
```diff
@@ -0,0 +1,32 @@
+package com.google.graphgeckos.dashboard.reader;
+
+public class BuildBotSetUpData {
+
+  // Name of the BuildBot
```
MForster, line 5: These comments should be JavaDoc.

src/main/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpData.java
```diff
@@ -0,0 +1,32 @@
...
+  public final long delay;
+
+  public BuildBotSetUpData(String name, long initialId, long delay) {
+    this.name = name;
+    this.initialId = initialId;
+    this.delay = delay;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
MForster, line 21: We'd normally use `AutoValue` for something like this: https://github.com/google/auto

src/main/resources/static/buildbots.txt
holounic: Done! https://github.com/googleinterns/step240-2020/wiki/BuildBotSetUpReader

MForster: Instead of inventing new file formats and parsers, we would usually just write a text proto file. 

https://github.com/protocolbuffers/protobuf

src/test/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpReaderTest.java
MForster: Using protos we wouldn't need that .

src/main/java/com/google/graphgeckos/dashboard/reader/BuildBotSetUpReader.java
```diff
@@ -0,0 +1,59 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.springframework.lang.NonNull;
+
+public class BuildBotSetUpReader {
+  /** Path to the config file with the initial BuildBot data. */
+  private static String filePath = "src/main/resources/static/buildbots.txt";
```
MForster, line 14: Don't use global mutable data.

MForster (changes requested):

src/main/resources/static/buildbots.txt
Somae: I agree. We should add a note about that and document it in the work that needs to be done. For the work in the internship using protos instead will be out of scope.

Somae:
Will merge for project completion. This will be added to a queue to be revisited later.

PR #199, opened 2020-09-23
Fix master test failures

MagoDelBlocco:
Fixes test failures related to exceptions thrown in the storage index methods. The most important bug was that there was no checking for a null associatedEntity in `registerNewBuildbot`, resulting in a cascade of NPEs.

PR #206, opened 2020-09-24
fix inverse logic in checkArgument

MagoDelBlocco:
Preconditions.checkArgument(CONDITION) checks for the `CONDITION` to be true. While merging #199 I mistakenly accepted previous changes, which had the CONDITION to be false under normal circumstances. This PR fixes that mistake and allows 2 tests to pass.


samuelsogbesan (approved):
oof LGTM

Somae (approved):
LGTM

PR #209, opened 2020-09-24
Fix test failure

MagoDelBlocco:
Accidentally removed condition to throw exception when setBuildbotIndex was called with an argument equal to the last previous known index.


samuelsogbesan (approved):
LGTM 😂

Somae (approved):

Somae:
I was surprised when you kept the equal sign last time, but figured it was intentional since you corrected the comment as well.

MagoDelBlocco:
I mistook that check for another one up above that was supposed to be without an equals sign, hence the confusion

PR #213, opened 2020-09-25
Catch undefined error when accessing builder array

samuelsogbesan:
This PR integrates a fix for an error returned when accessing the builders array from the API. It seems the undefined case wasn't being checked before accessing a builder's name field.


MagoDelBlocco (approved):
LGTM

krasimir-google (approved):
LGTM to merge this to fix the bug.
I suspect there may be a more conventional way in JS to do this, so I'd like you to reference this PR in a list of PRs that we'd like to get readability on later.

samuelsogbesan:
> LGTM to merge this to fix the bug.
> I suspect there may be a more conventional way in JS to do this, so I'd like you to reference this PR in a list of PRs that we'd like to get readability on later.

I'll create an issue listing them.

PR #215, opened 2020-09-25
add global revision status

MagoDelBlocco:
All builder statuses were not aggregated, and thus the front end received no info regarding the overall status of the build, required for color coding it and the status field on the snapshot. This PR adds a new datatype used for that, and the logic inside BuildInfo.

### DIFF
--- a/src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildInfo.java
+++ b/src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildInfo.java
@@ -70,6 +70,8 @@ public class BuildInfo {
   /**
    * The global compilation status of the revision.
    */
+  @Unindexed
+  @Field(name = "status")
   private RevisionStatus status;
 
   /**
@@ -87,7 +89,7 @@ public class BuildInfo {
     this.timestamp = Timestamp.parseTimestamp(creationData.getTimestamp());
     this.branch = creationData.getBranch();
     this.builders = new ArrayList<>();
-    this.status = analyseRevision();
+    this.status = RevisionStatus.lost;
   }
 
   @NonNull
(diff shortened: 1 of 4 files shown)


samuelsogbesan (approved):
LGTM, please add @Field annotation to status

MagoDelBlocco:
> LGTM, please add @field annotation to status

we don't mean to add this field in the datastore. If that were the case we would have needed a new component for analyzing this each time the entity is updated. I'll file an issue for that

samuelsogbesan:
> > LGTM, please add @field annotation to status
> 
> we don't mean to add this field in the datastore. If that were the case we would have needed a new component for analyzing this each time the entity is updated. I'll file an issue for that

Cool

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildInfo.java
```diff
@@ -124,6 +135,24 @@ public void addBuilder(@NonNull BuildBotData update) {
     builders.add(update);
   }
 
+  private RevisionStatus analyseRevision() {
+    boolean allPassed = false;
```
krasimir-google, line 139: I don't follow: the code below would mark `allPassed` to `true` if the builder statuses are [LOST, PASSED, LOST] for example. Maybe you need a different name for this?

I would be helpful to add a javadoc describing the cases where the different statuses get matched.

src/main/java/com/google/graphgeckos/dashboard/datatypes/RevisionStatus.java
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.graphgeckos.dashboard.datatypes;
+
+/**
+ * Represents global revision statuses.
+ */
+public enum RevisionStatus {
+  failed,
```
krasimir-google, line 21: turn those cases into uppercase, like in https://github.com/googleinterns/step240-2020/blob/master/src/main/java/com/google/graphgeckos/dashboard/datatypes/BuilderStatus.java

krasimir-google (changes requested):
Please add a couple of test cases that exercise the new ``analyzeRevision` functionality and status field.

src/main/java/com/google/graphgeckos/dashboard/datatypes/BuildInfo.java
MagoDelBlocco: I think the variable name is misleading. The status should indeed be true in the case you described, since it's the only valid datapoint that we have. I'll change the name and add a javadoc to the method

src/main/java/com/google/graphgeckos/dashboard/datatypes/RevisionStatus.java
MagoDelBlocco: This requires further rework from the front-end part. I think it would be best to leave them as they are now, as per @samuelsogbesan 's request, and these will be changed on both the front and back end parts in a future PR

krasimir-google: ok. Please add a TODO here

MagoDelBlocco: done

krasimir-google (approved):
Great! Thank you!

PR #216, opened 2020-09-25
Remove reanalyse call in BuildInfoTests

MagoDelBlocco:
The reanalyse call is integrated in the addBuilders method. This reanalyse method is private and triggers compilation failure everywhere it's placed. This PR removes all the calls from the tests.


samuelsogbesan (approved):
LGTM

krasimir-google (approved):
lgtm

PR #218, opened 2020-09-25
Fix error due to timestamp format mismatch

samuelsogbesan:
This PR fixes an error that arises due to formatting difference between how our timestamps are stored in our GCP datastore and how they are expected on the frontend. The frontend expected a string where it got a frontend, causing an error since React cannot render objects directly.


MagoDelBlocco (approved):
LGTM

krasimir-google (approved):
LGTM, please add it to the readability re-review list

PR #219, opened 2020-09-25
Scope variable properly

samuelsogbesan:
Scope variable properly so it can accessed inside component.


MagoDelBlocco (approved):
LGTM

krasimir-google (approved):

